PR #50, opened 2015-01-20
Make knowrob_vis webserver configurable.

mh0rst:
When using knowrob_vis, it is not always desirable to launch a webserver on a fixed port.
This makes the startup configurable on the parameter server like in this example launch file exerpt:
```
  <param name="initial_package" type="string" value="knowrob_vis" />
  <param name="initial_goal" type="string" value="visualisation_canvas" />
  <param name="knowrob_vis/webserver_start" type="bool" value="false" />
  <param name="knowrob_vis/webserver_port" type="bool" value="31337" />
  <node name="json_prolog" pkg="json_prolog" type="json_prolog_node" cwd="node" output="screen" />
```


mh0rst:
The commit 1694027297a4a73244ba6033bd82dc157daaa9be makes launching the webserver optional by refactoring it to a separate predicate, which makes this PR obsolete.

PR #63, opened 2015-06-18
Allow various levels of depth for classpath.txt file.

So-Cool:
For **collection of packages** the fixed path depth of the `classpath.txt` file works but when you develop just a **single package** you can expect the path to be one directory shorter e.g. `package_name/component_name/build/classpath.txt`
therefore, you won't have access to your Java classes from JPL in `rosprolog`.

With proposed here fix the path can be of arbitrary depth.


daniel86:
Thanks for the patch!

PR #87, opened 2016-03-10
 minor changes for semantic map loading from external programs

bbrieber:
This fix makes some queries on the semantic map much more flexible.

This should not break anything but it would be nice if someone tests this with the old semantic kitchen map.

@gaya- can you have a look at that?


gaya-:
@bbrieber Querying the old semantic map still works.

PR #88, opened 2016-03-15
added function transformToMatrix and new queries

yazdani:
for querying the new notation of the semantic map based on its changes to the older one

### DIFF
--- a/knowrob_common/knowrob_common/src/main/java/org/knowrob/utils/MathUtil.java
+++ b/knowrob_common/knowrob_common/src/main/java/org/knowrob/utils/MathUtil.java
@@ -40,6 +40,45 @@ import javax.vecmath.Quat4d;
 import javax.vecmath.Vector3d;
 
 public class MathUtil {
+
+    public static double[] transformToMatrix(float[] pose) {
+	    double[] m = new double[16];
+
+	    double xx = pose[3] * pose[3];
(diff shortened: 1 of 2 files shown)


knowrob_common/knowrob_common/src/main/java/org/knowrob/utils/MathUtil.java
```diff
@@ -40,6 +40,45 @@
 import javax.vecmath.Vector3d;
 
 public class MathUtil {
+
+    public static double[] transformToMatrix(float[] pose) {
```
daniel86: You can use Matrix4d, it supports to be initialized by quaternion/translation.
Also, there are already predicates `matrix_rotation` and `matrix_translation`,
they could be augmented so that when matrix is bound and rotation/translation is bound
then set the rotation/translation.

knowrob_objects/prolog/knowrob_objects.pl
```diff
@@ -34,11 +34,13 @@
       storagePlaceFor/2,
       storagePlaceForBecause/3,
       current_object_pose/2,
+      new_current_object_pose/2,
```
daniel86: Don't introduce new predicates. You can define rules with identical names but different
bodies.
Then, `current_object_pose` would support old-style and new-style

PR #93, opened 2016-04-28
fix

bbferka:
classifiers.pl does not exist anymore


daniel86:
@asilx I think you removed it when you did the WEKA integration

PR #98, opened 2016-05-03
added new queries to call the semantic map of kitchen with new notati…

yazdani:
These queries are built for the neweset version of the semantic map related to kitchen etc.
They are important for querying owl in CRAM


daniel86:
thanks for the commit!

PR #122, opened 2016-10-13
Added missing dependency of json_prolog.

airballking:
I encountered this missing dependency when compiling knowrob from source on a freshly installed Ubuntu 14.04.


daniel86:
thanks!

PR #128, opened 2016-11-01
Fix rosprolog-test: Only run tests for packages specified as arguments.

airballking:
A call to ```rosrun rosprolog rosprolog-test PKG-A,PKG-B``` would run tests for ```PKG-A, PKG-B```, and all their dependencies. 

To me, the testing of the dependencies seems like a bug because one could just add them as further packages to test. 

This pull request changes the behavior to only test ```PKG-A``` and ```PKG-B```, and not the dependencies.

### DIFF
--- a/rosprolog/scripts/rosprolog-test
+++ b/rosprolog/scripts/rosprolog-test
@@ -20,7 +20,7 @@ if [ -n "$1" ]; then
      fi
      IFS=','
    done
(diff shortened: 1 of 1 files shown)


airballking:
@daniel86 could you please have a look at this? Thanks!

airballking:
Thanks!

PR #133, opened 2016-12-12
[knowrob_map_tools] add conversion script from URDF to knowrob:SemanticEnvironmentMap

furushchev:
I'm not sure it's a right place to put these codes.
This pull request includes script to convert URDF model file into knowrob:SemanticEnvironmentMap.


furushchev:
@asilx Thanks for merging!

PR #143, opened 2017-06-21
Augment owl_has reasoning

mpomarlan:
with PropertyChain for object properties.

Example PropertyChain:

object property hP is defined in an ontology as a property chain of object properties uC o inverse hA (o is part of syntax here). We then assert rdf triples (s, uC, c), (p, hA, c) and nothing else. The property chain means that implicitly we also have (s, hP, p).

Old version: owl_has(s, hP, p) fails.

New version: owl_has(s, hP, p) succeeds.

The modifications to the owl_has predicate were tested. They also include some sanity checks (they will not follow circular property chains). In case of finding rdf triples (S, R, O) through a property chain, R and at least one of S or O must be bound.

### DIFF
--- a/knowrob_common/prolog/owl.pl
+++ b/knowrob_common/prolog/owl.pl
@@ -593,6 +593,84 @@ intersection_of(Nil, _) :-
 
 
 		 /*******************************
+		 *	  AUX PREDS (RD_FLIST)	*
+		 *******************************/
+
+%% Taken from semweb/rdf11. Needed to interpret PropertyChain axioms.
+
+kr_rdf_is_subject(T) :- atom(T).
+
+kr_rdf_list(RDFList, Prolog) :-
+    kr_rdf_is_subject(RDFList),
+    !,
+    kr_rdf_list_to_prolog(RDFList, Prolog).
+kr_rdf_list(RDFList, _Prolog) :-
+    type_error(rdf_subject, RDFList).
+
+kr_rdf_list_to_prolog(RDF, Prolog) :-
+    owl_same_as(RDF, 'http://www.w3.org/1999/02/22-rdf-syntax-ns#nil'),
+    !,
+    Prolog = [].
+kr_rdf_list_to_prolog(RDF, [H|T2]) :-
+    (   rdf_has(RDF, rdf:first, H0),
+        rdf_has(RDF, rdf:rest, T1)
+    *-> H = H0,
+        kr_rdf_list_to_prolog(T1, T2)
+    ;   type_error(rdf_list, RDF)
+    ).
+
+
+		 /*******************************
+		 *	  AUX PREDS (CIRC)	*
+		 *******************************/
+
+kr_circular_subproperty(SubProperty, SuperProperties) :- 
+        member(SuperProperty, SuperProperties),
+        owl_same_as(SubProperty, SuperProperty).
+
+kr_circular_subproperty(SubProperty, SuperProperties) :-
+        rdf_has(SubSubProperty, rdfs:subPropertyOf, SubProperty),
+        append(SuperProperties, [SubProperty], NewSuperProperties),
+        kr_circular_subproperty(SubSubProperty, NewSuperProperties).
+
+% An element in a property chain is either an object property or a description
+% containing an owl:inverseOf some object property.
+get_chain_property(Ch, Pr) :-
+        owl_individual_of(Ch, owl:'ObjectProperty'),
+        =(Ch, Pr).
+get_chain_property(Ch, Pr) :-
+        \+ owl_individual_of(Ch, owl:'ObjectProperty'),
+        rdf_has(Ch, owl:inverseOf, Pr).
+
+kr_get_props([], Props, Props).
+
+kr_get_props(PropChain, CrProps, Props) :-
+        nth0(0, PropChain, ChainFirst, RestChain),
+        get_chain_property(ChainFirst, Property),
+        append(CrProps, [Property], NewCrProps),
+        kr_get_props(RestChain, NewCrProps, Props).
+
+kr_get_props(PropChain, Props) :-
+        kr_get_props(PropChain, [], Props).
+
+kr_circular_chain(ChainProperties, PreviousProperties) :-
+        member(P, PreviousProperties),
+        member(P, ChainProperties).
+
+kr_circular_chain(ChainProperties, PreviousProperties) :-
+        member(P, PreviousProperties),
+        rdf_has(P, owl:inverseOf, IP),
+        member(IP, ChainProperties).
+
+kr_circular_chain(ChainProperties, PreviousProperties) :-
+        member(CP, ChainProperties),
+        rdf_has(CP, owl:propertyChainAxiom, RDFList),
+        kr_rdf_list(RDFList, CPChain),
+        kr_get_props(CPChain, CPProps),
+        append(PreviousProperties, ChainProperties, NewPreviousProperties),
+        kr_circular_chain(CPProps, NewPreviousProperties).
+
+		 /*******************************
 		 *	  OWL PROPERTIES	*
 		 *******************************/
 
@@ -682,20 +760,23 @@ owl_same_as(X, Y, Visited) :-
 %	@bug	owl_has_direct/3 also uses SWRL rules.  This should be
 %		moved elsewhere.
 
+%%      MP 06/2017: also add support for reasoning based on subPropertyOf and PropertyChain axioms of object properties
+%%	TODO: add support for Equivalent properties and mixes of sub properties/Equivalent ones
+%%      TODO: add support for subPropertyOf for data properties
+
 owl_has_direct(S, P, O) :-
-	rdf(S, P, O).
-owl_has_direct(S, P, O) :-
-	(   rdf_has(P, owl:inverseOf, P2)
-	->  true
-	;   rdf_has(P2, owl:inverseOf, P)
-	),
-	rdf_has(O, P2, S).		% TBD: must call owl_has_direct/3
+        owl_has_direct_internal(S, P, O).
+
 owl_has_direct(S, P, O) :-
 	rdfs_individual_of(P, owl:'SymmetricProperty'),
-	rdf(O, P, S).
-owl_has_direct(S, P, O) :-
-	owl_use_has_value(S, P, O).
+	owl_has_direct_internal(O, P, S).
 
+owl_has_direct(S, P, O) :-
+        (   rdf_has(P, owl:inverseOf, P2)
+        ->  true
+        ;   rdf_has(P2, owl:inverseOf, P)
+        ),
+        owl_has_direct_internal(O, P2, S).
 
 %----------------------------------------------------------
 % added by BJW for use of OWL with SWRL rules, highly experimental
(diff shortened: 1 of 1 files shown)


knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+    ).
+
+
+		 /*******************************
+		 *	  AUX PREDS (CIRC)	*
+		 *******************************/
+
+% An element in a property chain is either an object property or a description
+% containing an owl:inverseOf some object property.
+get_chain_property(Ch, Pr) :-
```
daniel86: rename to `owl_chain_property` or so. Since usually there are no "getter" functions in Prolog. It all depends on which variable is bound or unbound

knowrob_common/prolog/owl.pl
```diff
@@ -682,20 +751,30 @@
...
+
 owl_has_direct(S, P, O) :-
-	owl_use_has_value(S, P, O).
+        (   rdf_has(P, owl:inverseOf, P2)
+        ->  true
+        ;   rdf_has(P2, owl:inverseOf, P)
+        ),
+        owl_has_direct_internal(O, P2, S).
 
+owl_has_direct(S, P, O) :-
```
daniel86: Could be merged with above clause to avoid redundant if statement

knowrob_common/prolog/owl.pl
```diff
@@ -709,6 +788,79 @@
...
+	rdf_has(S, P, O).
+
+%% If P is bound to an object property, see if any of its PropertyChain axioms is able to produce explicitly known triples.
+%% ASSUMPTION: no circular PropertyChain axioms (example, P defined as A o B and A defined as P o B)
+owl_has_direct_internal(S, P, O) :-
+        nonvar(P),
+        owl_individual_of(P, owl:'ObjectProperty'),
+        rdf_has(P, owl:propertyChainAxiom, RDFList),
+        kr_rdf_list(RDFList, PropChain),
+% We need this because the PropChain contains a list of either property or Description individuals (where a Description individual refers to an inverse property),
```
daniel86: Properties are also RDF descriptions

knowrob_common/prolog/owl.pl
```diff
@@ -709,6 +788,79 @@
...
+
+owl_has_property_chain_S2O(S, PropChain, O) :-
+        nth0(0, PropChain, P, RestChain),
+        owl_individual_of(P, owl:'ObjectProperty'),
+        owl_has(S, P, Oi),
+        owl_has_property_chain_S2O(Oi, RestChain, O).
+
+owl_has_property_chain_S2O(S, PropChain, O) :-
+        nth0(0, PropChain, P, RestChain),
+        owl_has(P, owl:inverseOf, PI),
```
daniel86: Why handle inverse here again? `owl_has` handles inverse already.
Just calling `owl_has(S,P,Oi)` should be enough, or?
Then we would need only above clause and can remove this one.

knowrob_common/prolog/owl.pl
```diff
@@ -709,6 +788,79 @@
...
+
+owl_has_property_chain_O2S(O, PropChainRev, S) :-
+        nth0(0, PropChainRev, P, RestChain),
+        owl_individual_of(P, owl:'ObjectProperty'),
+        owl_has(Si, P, O),
+        owl_has_property_chain_O2S(Si, RestChain, S).
+
+owl_has_property_chain_O2S(O, PropChainRev, S) :-
+        nth0(0, PropChainRev, P, RestChain),
+        owl_has(P, owl:inverseOf, PI),
```
daniel86: Same comment as above: `owl_has` already handles inverse relations

knowrob_common/prolog/owl.pl
```diff
@@ -709,6 +788,79 @@
...
+
+owl_has_property_chain_S2O(S, PropChain, O) :-
+        nth0(0, PropChain, P, RestChain),
+        owl_has(P, owl:inverseOf, PI),
+        owl_has(Oi, PI, S),
+        owl_has_property_chain_S2O(Oi, RestChain, O).
+
+owl_has_property_chain_O2S(S, [], S).
+
+owl_has_property_chain_O2S(O, PropChainRev, S) :-
```
daniel86: For me it seems that `owl_has_property_chain_O2S` can be removed entirely.
Because it should not matter in which direction we check the chain since we take into account inverse relations. Thus, `owl_has_property_chain_S2O` should be sufficient.

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
 
 		 /*******************************
+		 *	  AUX PREDS (RD_FLIST)	*
+		 *******************************/
+
+%% Taken from semweb/rdf11. Needed to interpret PropertyChain axioms.
+
+kr_rdf_is_subject(T) :- atom(T).
+
+kr_rdf_list(RDFList, Prolog) :-
```
daniel86: There is a predicate `rdfs_list_to_prolog_list`. Could you use this instead?

mpomarlan: Sue.

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+    ;   type_error(rdf_list, RDF)
+    ).
+
+
+		 /*******************************
+		 *	  AUX PREDS (CIRC)	*
+		 *******************************/
+
+% An element in a property chain is either an object property or a description
+% containing an owl:inverseOf some object property.
```
mpomarlan: This check is needed because the RDFList returned when querying for PropertyChains may contain either object properties or 'anonymous' properties that are not individuals of owl:ObjectProperty, but do have an owl:inverseOf relation to an individual of type owl:'ObjectProperty'.

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+get_chain_property(Ch, Pr) :-
+        owl_individual_of(Ch, owl:'ObjectProperty'),
+        =(Ch, Pr).
```
mpomarlan: Done.

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+get_chain_property(Ch, Pr) :-
+        \+ owl_individual_of(Ch, owl:'ObjectProperty'),
+        rdf_has(Ch, owl:inverseOf, Pr).
+
+kr_get_props([], Props, Props).
+
+kr_get_props(PropChain, CrProps, Props) :-
```
mpomarlan: Done.

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+        nth0(0, PropChain, ChainFirst, RestChain),
```
mpomarlan: Done.

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+        get_chain_property(ChainFirst, Property),
+        append(CrProps, [Property], NewCrProps),
```
mpomarlan: Done

knowrob_common/prolog/owl.pl
```diff
@@ -593,6 +593,75 @@
...
+        kr_get_props(RestChain, NewCrProps, Props).
```
mpomarlan: Done.

knowrob_common/prolog/owl.pl
mpomarlan: Done.

knowrob_common/prolog/owl.pl
mpomarlan: Done.

knowrob_common/prolog/owl.pl
mpomarlan: In this case, the problem is that the RDFList may contain either individuals of type owl:'ObjectProperty', or individuals who are not object properties but have owl:inverseOf some owl:'ObjectProperty'.

knowrob_common/prolog/owl.pl
```diff
@@ -709,6 +788,79 @@
...
+
+owl_has_property_chain(S, PropChain, O) :-
+        var(S),
+        nonvar(O),
+        reverse(PropChain, PropChainRev),
+        owl_has_property_chain_O2S(O, PropChainRev, S).
+
+owl_has_property_chain_S2O(O, [], O).
+
+owl_has_property_chain_S2O(S, PropChain, O) :-
```
mpomarlan: Done

knowrob_common/prolog/owl.pl
mpomarlan: Done.

knowrob_common/prolog/owl.pl
mpomarlan: Maybe, but I'd rather not risk it. the *_S2O and *_O2S are written with the assumption (checked with var/nonvar predicates) that one of the S, O variables is bound, and the functions will start verifying the property chain from the end that is bound. So for example, if S is bound, then *_S2O is called, and it will check the chain starting from S. While it may be possible to start verifying the chain from the unbound end, I imagine it would be very inefficient as many branches will be explored before one is found that completes the chain to the bound end.

knowrob_common/prolog/owl.pl
mpomarlan: Done.

knowrob_common/prolog/owl.pl
daniel86: I didn't suggest to start with unbound var. Do a check before calling `owl_has_property_chain` if S unbound switch S and O and generate inverse property chain for checking.

knowrob_common/prolog/owl.pl
daniel86: `owl_has(S,P,Oi)` should hold when `owl_has(Oi,Pi,S)` holds with `Pi = inverse_of(P)`

knowrob_common/prolog/owl.pl
daniel86: For me it seems semantically wrong that these inverse property descriptions are not object properties themselves

knowrob_common/prolog/owl.pl
daniel86: Could you take a look why there are no results for anonymous inverse properties in `owl_has`?
It's not obvious to me. In `owl_has_direct` the inverse is handled, and I don't see why it shouldn't work for unnamed inverse properties.

knowrob_common/prolog/owl.pl
mpomarlan: Does this save us any code though?

knowrob_common/prolog/owl.pl
mpomarlan: Simple. rdf_has fails on "unnamed inverse properties".

mpomarlan: Tested a code version with the last clauses for *_S2O and *_O2S removed. My 'unit tests' still work, so I've committed it to this fork.

knowrob_common/prolog/owl.pl
mpomarlan: Also, is there any inverse_of predicate that can, if necessary, create a new individual to be the inverse?

knowrob_common/prolog/owl.pl
daniel86: `rdf_has` fails for sure because RDF does not have inverse_of semantics, OWL adds this

knowrob_common/prolog/owl.pl
```diff
@@ -709,6 +762,64 @@
...
+owl_has_property_chain_S2O(O, [], O).
+
+owl_has_property_chain_S2O(S, [P|RestChain], O) :-
+        owl_has(S, P, Oi),
+        owl_has_property_chain_S2O(Oi, RestChain, O).
+
+owl_has_property_chain_O2S(S, [], S).
+
+owl_has_property_chain_O2S(O, [P|RestChain], S) :-
+        owl_has(Si, P, O),
```
daniel86: Shouldn't that be `owl_has(O,P,Si)` instead?
Since `P` is inverse relation with `O` as subject.
Then you can remove this clause and just use the above one for both cases.

mpomarlan: No.

Currently P is not the inverse of anything. It is a property extracted from the PropertyChain as-is. Where is that inverse_of predicate btw?

daniel86: Ok got it, was overseeing that you just reverse the list without replacing it with inverse relation

PR #154, opened 2017-08-25
json_prolog now does not return empty lists as atoms

ichumuh:


asilx:
@ichumuh  @daniel86 
isListNil() function is introduced in JPL 7. Are you sure that we want to remove older versions of JPL compability?

ichumuh:
Sorry, I didn't realize that this function is not available in older versions. In the kinetic branch we need jpl7 anyway. But I guess its not difficult to do it without this function for the master branch.

PR #157, opened 2017-08-28
fixed compile error

ichumuh:
This should fix the compile error. However, it does not completely fix the bug I was trying to fix with my initial pull request.
With this version I get the following results:
query: A=[]
result: []
query: A='[]'
result: [] <-- this should be an atom
It is not so easy to fix this properly because I couldn't find good jpl3 documentation, maybe we should just switch to jpl7 for the master branch as well?


daniel86:
Well as stated before [] is an atom!
In SWI Prolog, do this:
?- X = '[]'.
X = [].

The single quotes are only used to enforce atoms, but [] is an atom anyway.
Also this holds:
?- [] = '[]'.
true.

So this result is expected.

daniel86:
On the other hand, not sure about the design decision to make [] an atom.

Maybe (somehow) related to this behaviour is this:

    ?- [a] = .(a,[]).
    true.

Where .(Head,Rest) is internal list representation of Prolog
(brackets are only syntax sugar again).

PR #161, opened 2017-09-11
json_prolog_commandline update from kinetic

ichumuh:
added tabcompletion, namespace loading, limited history and dont print all solutions at once to json_prolog_commandline.


daniel86:
very cool, thanks a lot!

PR #167, opened 2017-11-24
fix deadlock

bbferka:
- changed synchronizations to break deadlock in cases where json_prolog query triggers another query;
- added initial_package as an argument to the launch file, allowing the change of this from the terminal

### DIFF
--- a/json_prolog/launch/json_prolog.launch
+++ b/json_prolog/launch/json_prolog.launch
@@ -1,4 +1,5 @@
 <launch>
-  <param name="initial_package" type="string" value="knowrob_common" />
+  <arg name="initial_package" default="knowrob_common" />
+  <param name="initial_package" type="string" value="$(arg initial_package)" />
   <node name="json_prolog" pkg="json_prolog" type="json_prolog_node" cwd="node" output="screen" />
 </launch>
--- a/json_prolog/json_prolog/src/main/java/org/knowrob/json_prolog/JSONPrologNode.java
+++ b/json_prolog/json_prolog/src/main/java/org/knowrob/json_prolog/JSONPrologNode.java
@@ -151,7 +151,7 @@ public class JSONPrologNode extends AbstractNodeMain {
 			//	response.setMessage("Failed to close incremental query.");
 			//	return;
 			//}
-			
+		        	
 			if ( queries.get(request.getId()) != null ) {
 				response.setOk(false);
 				response.setMessage("Already processing a query with id " + request.getId());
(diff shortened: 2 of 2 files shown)


bbferka:
@daniel86 we only looked at SimpleQuery, maybe the normal Query also has this issue?

daniel86:
@bbferka thanks for the pull request! I'll merge after symposium ;)

Well actually I never understood what the difference is. What makes simple query simple?
I think no one uses the query service, only simple_query is used.
But yes, might be affected as well.

cheers!

json_prolog/launch/json_prolog.launch
```diff
@@ -1,4 +1,5 @@
 <launch>
-  <param name="initial_package" type="string" value="knowrob_common" />
+  <arg name="initial_package" default="knowrob_common" />
+  <param name="initial_package" type="string" value="$(arg initial_package)" />
```
daniel86: why did you changed this?

bbferka: It's an enhancement. It allows setting the initial package from the command line when executing the launch file.

mh0rst:
This fix might lead to undefined behaviour.
The prolog interpreter behind json_prolog supports only one active query at a time (see http://jpl7.org/740/JavaApiOverview.jsp#Multi-Threaded%20Queries and the "Terminating Queries" chapter above).
I recommend reverting this (excluding the launch file improvement) and refactoring the dead-locking query.

daniel86:
what a bummer :/ thanks for pointing that out @mh0rst 
I will take a look at this again...

bbferka:
Yea, That sux bigtime!  I was afraid of something like this. We did not check every implication of this change when we did it. If you think it can cause serious problems it can be reverted, I think I can just use a different fix for this problem. It is strange however that the prolog interpreter works for me with two queries (basically in one call initiates the other one and waits until the 2nd call terminates);

daniel86:
well we can't conclude from that that it works in general.
but ending up in a deadlock is also not good behaviour.

I find it a bit confusing that the jpl Query objects explicitly
support threading: Each query is bound to a thread
and it is also stated that each thread has it's own Prolog engine
attached -- however, not sure if this is already the way to avoid
mentioned multithreading problems.

@mh0rst you were running into problems?

mh0rst:
No problems on my side so far, I don't think openease sends multiple queries concurrently anywhere.
I just noticed this commit while trying to hunt some other bugs and checked if that was causing one of the issues I experienced (it did not).

My objection is purely based on the (extremely outdated and inaccurate) documentation of jpl and some commits in the https://github.com/SWI-Prolog/packages-jpl repository. From version 3.1.4-alpha (which indigo currently uses) up to 7.7.5 the developers touched thread safety in both Java and C parts.
Due to the complexity of swipl and jpl, it is hard to see if thread safety is guaranteed in every circumstance with the current versions we are using. I would at least run some multithreaded tests against json_prolog to see if it works with many clients firing at the node.

PR #182, opened 2018-01-11
Fixes for Kinetic on 16.04

jworch:


knowrob_common/knowrob_common/src/main/java/org/knowrob/prolog/PrologInterface.java
```diff
@@ -62,17 +64,17 @@ public static void initJPLProlog(String initPackage) {
...
     			args.add( "-nosignals" );
 
     			String rosprolog = RosUtilities.rospackFind("rosprolog");
-    			jpl.fli.Prolog.set_default_init_args( args.toArray( new String[0] ) );
+    			org.jpl7.fli.Prolog.set_default_init_args( args.toArray( new String[0] ) );
 
     			// load the appropriate startup file for this context
-    			new jpl.Query("ensure_loaded('"+rosprolog+"/prolog/init.pl"+"')").oneSolution();
-    			new jpl.Query("register_ros_package('"+initPackage+"')").oneSolution();
+    			new org.jpl7.Query("ensure_loaded('"+rosprolog+"/prolog/init.pl"+"')").oneSolution();
```
daniel86: can't you use just Query here? org.jpl7.Query imported at the top.
same for some other statements.

knowrob_common/knowrob_common/src/main/java/org/knowrob/prolog/PrologInterface.java
```diff
@@ -62,17 +64,17 @@ public static void initJPLProlog(String initPackage) {
     	if(!initialized) {
 
     		try {
-    			Vector<String> args= new Vector<String>(Arrays.asList(jpl.fli.Prolog.get_default_init_args()));
+    			Vector<String> args= new Vector<String>(Arrays.asList(org.jpl7.fli.Prolog.get_default_init_args()));
```
daniel86: Could you import org.jpl7.fli.Prolog, and then use just Prolog.
I guess this could be good when a new JPL version is there
to avoid merge conflicts between branches

knowrob_common/knowrob_common/src/main/java/org/knowrob/prolog/PrologInterface.java
```diff
@@ -140,7 +141,7 @@ public static void initLocalProlog(String initFile) {
     		// Build the result
     		for (int i=0; i<solutions.length; i++) {
     			@SuppressWarnings("rawtypes")
-				Hashtable solution = solutions[i];
+				Map<String, Term> solution = solutions[i];
```
daniel86: is the SuppressWarnings still needed?

knowrob_srdl/knowrob_srdl/build.gradle
```diff
@@ -19,7 +19,7 @@ apply plugin: 'application'
 mainClassName = "org.knowrob.knowrob_srdl.URDF2SRDL"
 
 dependencies {
-  compile 'org.ros.rosjava_core:rosjava:[0.1,)'
+  compile 'org.ros.rosjava_core:rosjava:[0.3,0.4)'
```
daniel86: I wonder a bit, how do you come up with these version numbers?

knowrob_common/prolog/plunit.pl
```diff
@@ -1,1531 +0,0 @@
-/*  $Id$
```
daniel86: this should not be part of the commit. Could you please restore the file for this pull request? This will result in merge conflict when I merge master in kinetic branch.

knowrob_common/knowrob_common/src/main/java/org/knowrob/prolog/PrologInterface.java
jworch: Should be okay to remove it, since Map is typed.

knowrob_common/prolog/plunit.pl
jworch: Don't know how that happened -.-

knowrob_srdl/knowrob_srdl/build.gradle
jworch: By locate rosjava_core on my machine, I found rosjava core 0.3.5
You can either specify the exact version [0.3.5], or allow everything >= 0.3 (using [0.3.+]), or allow version numbers between 0.3 and 0.4.
I was going for the latter, because I expect minor version changes to be not api breaking. Could also be set to 0.3.5 if you prefer.

daniel86: sounds reasonable

knowrob_common/prolog/plunit.pl
jworch: But wait. Back then you told me to remove that file... Still want that to be restored?

daniel86:
thanks for your help!

PR #190, opened 2018-01-31
Prolog-wrapper for the C++ URDF parser.

airballking:
@daniel86 as discussed last week, here comes the prolog URDF parser. It covers all documented elements of links and joints of an URDF robot. Unit tests and doc-strings are also there.

PR #191, opened 2018-02-06
Kinetic urdf parsing

airballking:
@daniel86 this is the URDF parser in prolog. Developed under ROS Kinetic. Hence, the pull request into the kinetic branch. This replaces the old pull request #190


airballking:
@daniel86 could you please have a look at this?

daniel86:
@airballking yes sure, already looked at it a bit, give me more time. It's not urgent or am I wrong?

airballking:
You're right, it is not urgent. Thanks for looking into it!

knowrob_srdl/prolog/knowrob/urdf_parser.pl
```diff
@@ -0,0 +1,357 @@
...
+:- module(urdf_parser,
+    [
+      load_urdf_file/1,
+      load_urdf_param/1,
+      load_urdf_string/1,
+      robot_name/1,
+      root_link_name/1,
+      link_names/1,
+      link_parent_joint/2,
+      link_child_joints/2,
```
daniel86: potentially can be joined with `link_parent_joint` -- Prolog predicates can work in both directions!

knowrob_srdl/prolog/knowrob/urdf_parser.pl
```diff
@@ -0,0 +1,357 @@
...
+      link_material_texture/3,
+      link_num_collisions/2,
+      link_collision_type/3,
+      link_collision_name/3,
+      link_collision_origin/3,
+      link_collision_geometry/3,
+      joint_names/1,
+      joint_type/2,
+      joint_child_link/2,
+      joint_parent_link/2,
```
daniel86: parent-child relation can be covered by single predicate.

knowrob_srdl/prolog/knowrob/urdf_parser.pl
```diff
@@ -0,0 +1,357 @@
...
+      joint_axis/2,
+      joint_origin/2,
+      joint_lower_pos_limit/2,
+      joint_upper_pos_limit/2,
```
daniel86: I would suggest `joint_pos_limit(Joint, [Lower,Upper])`

knowrob_srdl/prolog/knowrob/urdf_parser.pl
```diff
@@ -0,0 +1,357 @@
...
+      joint_dynamics_damping/2,
+      joint_dynamics_friction/2,
+      joint_mimic_joint_name/2,
+      joint_mimic_multiplier/2,
+      joint_mimic_offset/2,
+      joint_safety_lower_limit/2,
+      joint_safety_upper_limit/2,
+      joint_safety_kp/2,
+      joint_safety_kv/2
+    ]).
```
daniel86: as pointed out earlier, some of the rules could be joined. But as long as these predicates are only used internally, I don't mind much

knowrob_srdl/prolog/knowrob/urdf_parser.plt
```diff
@@ -0,0 +1,400 @@
...
+
+test(robot_name_pr2) :-
+  robot_name(pr2).
+
+test(root_link_name_pr2) :-
+  root_link_name(base_footprint).
+
+test(joint_names_pr2) :-
+  joint_names(Names),
+  Names == [base_bellow_joint, base_footprint_joint, base_laser_joint,
```
daniel86: what the heck is github doing here? can't really review this file through github

knowrob_srdl/src/urdf_parser.cpp
```diff
@@ -0,0 +1,852 @@
...
+            dim_term[0] = mesh->filename.c_str();
+            PlTail scale(dim_term[1]);
+            scale.append(mesh->scale.x);
+            scale.append(mesh->scale.y);
+            scale.append(mesh->scale.z);
+            scale.close();
+            return PlCompound("mesh", dim_term);
+        }
+        default:
+            throw std::runtime_error("Encountered unsupported geometry type.");
```
daniel86: what happens if you throw exception and not handle it in C++?

knowrob_srdl/src/urdf_parser.cpp
```diff
@@ -0,0 +1,852 @@
...
+        long index = (long) PL_A2;
+        urdf::LinkConstSharedPtr link = get_link(link_name);
+        if (!link_has_visual_with_index(link, index) ||
+                !link->visual_array[index]->material)
+            return false;
+        PL_A3 = to_prolog_rgba(link->visual_array[index]->material->color);
+        return true;
+    } catch (const std::runtime_error& e) {
+        ROS_ERROR("%s", e.what());
+        return false;
```
daniel86: I guess this causes the call to `fail` ?
Also when does this happen? runtime_error is not very specific.

knowrob_srdl/src/urdf_parser.cpp
```diff
@@ -0,0 +1,852 @@
...
+
+PREDICATE(link_collision_type, 3) {
+    try {
+        std::string link_name((char*) PL_A1);
+        long index = (long) PL_A2;
+        urdf::LinkConstSharedPtr link = get_link(link_name);
+        if (!link_has_collision_with_index(link, index) ||
+                !link->collision_array[index]->geometry)
+            return false;
+        switch (link->collision_array[index]->geometry->type) {
```
daniel86: I prefer to use static arrays indexed by the type id (supposedly starting at 0), would be a few lines less :)
but surely just cosmetic

knowrob_srdl/src/urdf_parser.cpp
```diff
@@ -0,0 +1,852 @@
...
+        ROS_ERROR("%s", e.what());
+        return false;
+    }
+}
+
+PREDICATE(link_collision_name, 3) {
+    try {
+        std::string link_name((char*) PL_A1);
+        long index = (long) PL_A2;
+        urdf::LinkConstSharedPtr link = get_link(link_name);
```
daniel86: the shared ptr has some reference counting inside and auto deletes if counter reaches zero I guess?

knowrob_srdl/src/urdf_parser.cpp
```diff
@@ -0,0 +1,852 @@
...
+}
+
+/**************************************/
+/******** JOINT PROPERTIES ************/
+/**************************************/
+
+urdf::JointConstSharedPtr get_joint(const std::string& joint_name) {
+    urdf::JointConstSharedPtr joint = get_robot_model()->getJoint(joint_name);
+    if (!joint)
+        throw std::runtime_error("No joint with name '" + joint_name + "' in parsed URDF.");
```
daniel86: could we use more specific exception types?

knowrob_srdl/src/urdf_parser.cpp
```diff
@@ -0,0 +1,852 @@
...
+bool joint_has_effort_limit(urdf::JointConstSharedPtr joint) {
+    // joints that have velocity limits should also have effort limits
+    return joint_has_vel_limit(joint);
+}
+
+
+PREDICATE(joint_names, 1) {
+    try {
+        PlTail names(PL_A1);
+        for (auto const& joint_entry: get_robot_model()->joints_)
```
daniel86: never seen this keyword... so the type is inferred by the compiler, or what does it mean?

PR #192, opened 2018-02-07
Fixes for openEASE and added SRDL for Pepper

sasjonge:
Ths pull request will add an SRDL File for Pepper and fixed some problems for  openEASE.


daniel86:
danke!

PR #220, opened 2019-08-19
Changes from daniel86 fork & additional memory predicates

seba90:
Includes the changes from daniel86 fork.
In addition, the predicates mem_event_begin and mem_event_end include a one parameter version. This allows to create the start and end time of an event, without the need to add the current time stamp as an additional parameter.


daniel86:
Hi seba! Thanks for the request. I want to do some more testing before I merge my fork back to knowrob master. Please let me do this before, or just cherry-pick your commits with the knowrob master and do a pull request in case this does not create conflicts, or do a pull request against my fork, then these will be merged the next time I sync between my fork and the main repo.

PR #232, opened 2020-04-14
feat(ros_memory.pl): Add rosnode kill to achieve a more stable shutdown of the ros_logger.

seba90:


daniel86:
the TF logger code is now part of knowrob

PR #244, opened 2020-08-31
Moved openease_knowrob into knowrob repository and added settings to use read only remote mongo db

sasjonge:


src/db/mongo/client.pl
```diff
@@ -129,6 +131,22 @@
...
+% Make sure to call *mng_cursor_destroy/1* once
+% you are done querying.
+%
+% @param DB The database name
+% @param Collection The collection name
+% @param Cursor The id of a new mongo DB cursor
+%
+mng_cursor_create(DB, Collection, Cursor) :-
+  ( setting(mng_client:tripledb_read_only, true)
+    -> true
```
daniel86, line 146: this is not ok.

Cursors must also be created in read-only case, they are used for querying!

src/db/mongo/client.pl
```diff
@@ -34,7 +36,7 @@
 :- dynamic mng_db_name/1.
 
 % define some settings
-:- setting(db_name, atom, roslog,
+:- setting(db_name, atom, 'roslog',
```
daniel86, line 39: single quotes not needed here.
Try yourself `a = 'a'` holds

src/db/mongo/client.pl
```diff
@@ -7,13 +7,15 @@
...
       mng_index_create/2,
+      mng_index_create/3,
       mng_export/1,
       mng_export_collection/2,
       mng_dump/2,
       mng_dump_collection/3,
       mng_restore/2,
       mng_regex_prefix/2,
       mng_cursor_create/3,
+      mng_cursor_create_core/3,
```
daniel86, line 18: should not be exported

src/db/mongo/MongoInterface.cpp
```diff
@@ -81,10 +81,15 @@ MongoInterface::MongoInterface()
 	bson_error_t err;
 	std::string mongo_uri;
 	if(!rosprolog_kb::node().getParam(std::string("mongodb_uri"),mongo_uri)) {
-		char *mongo_host = std::getenv("MONGO_PORT_27017_TCP_ADDR");
+		char *mongo_host = std::getenv("NEEMHUB_MONGO_HOST");
 		if(mongo_host != NULL) {
+			char *mongo_user_name = std::getenv("NEEMHUB_MONGO_USER");
```
daniel86, line 86: better: KNOWROB_MONGO_USER and so on

docker/Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+# Initialize the catkin workspace
+USER ros
+WORKDIR /home/ros/src
+RUN /usr/bin/python /opt/ros/melodic/bin/catkin_init_workspace
+
+RUN if [ "x$LOCAL_BUILD" = "x0" ] ; then \
+    rm -rf /home/ros/src/* && \
+    git clone https://github.com/sasjonge/knowrob.git && \
+    git clone https://github.com/knowrob/rosprolog.git && \
+    git clone https://github.com/knowrob/rosowl.git && \
```
daniel86, line 86: not needed

docker/Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+    git clone https://github.com/knowrob/genowl.git && \
```
daniel86, line 87: not needed

daniel86 (changes requested):

src/db/mongo/MongoInterface.cpp
sasjonge: I tried to keep the naming consistent with the names in the openease packages. Should we rather use KNOWROB_MONGO_USER everywhere or NEEMHUB_MONGO_USER everywhere?

sasjonge: Other possibility would be REMOTE_MONGO_USER

daniel86: it might be most useful to have a "KNOWROB_" prefix.
First, then it is obvious to which project this belongs when looking at the env.
Second, there could be also other infrastructure then NEEM-Hub used as a backend for KnowRob.
Third, the env is global and we need to avoid name clashes so I would vote against "REMOTE_" prefix

settings/default.pl
```diff
@@ -11,3 +11,12 @@
 
 %	Mongo DB
 setting(mng_client:db_name, roslog).
+
+% Flag for read only mongo databases
+setting(mng_client:read_only, false).
+
+%	Remote Mongo Setting
+% setting(mng_client:db_name, 'neems').
```
daniel86, line 19: remove line

launch/openease.launch
```diff
@@ -0,0 +1,22 @@
+<launch>
+  <!-- Configure settings file. -->
+  <arg name="knowrob_settings" default="$(find knowrob)/settings/neemhub.pl" />
+  <env name="KNOWROB_SETTINGS" value="$(arg knowrob_settings)" />
+
+  <!-- param name="/ros_mac_authentication/secret_file_location" value="/etc/rosauth/secret" -->
```
daniel86, line 6: remove line and below

settings/neemhub.pl
```diff
@@ -0,0 +1,19 @@
+
+%	Plugins
+setting(knowrob:plugins,
+	[ 'comm/ros/tf/tf_plugin',
+	  'comm/ros/marker/marker_plugin'
+	]).
+
+%	Triple DB
+setting(tripledb:path, 'db/mongo/tripledb/plugin').
+setting(tripledb:drop_graphs, []). % erase "user" graph on start-up
```
daniel86, line 10: comment is not appropriate

src/db/mongo/MongoInterface.cpp
```diff
@@ -81,10 +81,15 @@ MongoInterface::MongoInterface()
 	bson_error_t err;
 	std::string mongo_uri;
 	if(!rosprolog_kb::node().getParam(std::string("mongodb_uri"),mongo_uri)) {
-		char *mongo_host = std::getenv("MONGO_PORT_27017_TCP_ADDR");
+		char *mongo_host = std::getenv("KNOWROB_MONGO_HOST");
 		if(mongo_host != NULL) {
+			char *mongo_user_name = std::getenv("KNOWROB_MONGO_USER");
+			char *mongo_user_pw = std::getenv("KNOWROB_MONGO_PASS");
```
daniel86, line 87: PW should be optional

src/db/mongo/MongoInterface.cpp
```diff
@@ -81,10 +81,15 @@ MongoInterface::MongoInterface()
...
+			char *mongo_database = std::getenv("KNOWROB_MONGO_DB");
+			char *mongo_port = std::getenv("KNOWROB_MONGO_PORT");
```
daniel86, line 89: port should be optional (mongo has a default port)

src/db/mongo/MongoInterface.cpp
```diff
@@ -81,10 +81,15 @@ MongoInterface::MongoInterface()
 	bson_error_t err;
 	std::string mongo_uri;
 	if(!rosprolog_kb::node().getParam(std::string("mongodb_uri"),mongo_uri)) {
-		char *mongo_host = std::getenv("MONGO_PORT_27017_TCP_ADDR");
+		char *mongo_host = std::getenv("KNOWROB_MONGO_HOST");
 		if(mongo_host != NULL) {
+			char *mongo_user_name = std::getenv("KNOWROB_MONGO_USER");
+			char *mongo_user_pw = std::getenv("KNOWROB_MONGO_PASS");
+			char *mongo_database = std::getenv("KNOWROB_MONGO_DB");
```
daniel86, line 88: but there is a setting to configure this *mng_client:db_name*, can't we use this instead of env?

src/db/mongo/MongoInterface.cpp
```diff
@@ -81,10 +81,15 @@ MongoInterface::MongoInterface()
 	bson_error_t err;
 	std::string mongo_uri;
 	if(!rosprolog_kb::node().getParam(std::string("mongodb_uri"),mongo_uri)) {
-		char *mongo_host = std::getenv("MONGO_PORT_27017_TCP_ADDR");
+		char *mongo_host = std::getenv("KNOWROB_MONGO_HOST");
 		if(mongo_host != NULL) {
+			char *mongo_user_name = std::getenv("KNOWROB_MONGO_USER");
```
daniel86, line 86: user should be optional

src/db/mongo/client.pl
```diff
@@ -40,6 +42,14 @@
 :- setting(mng_client:db_name, DBName),
    assertz(mng_db_name(DBName)).
 
+% Get db and collection for this type of data, e.g. triples
```
daniel86, line 45: comments of exported predicates must use pldoc syntax

daniel86 (changes requested):

src/db/mongo/MongoInterface.cpp
daniel86: why to append the DB name anyway? I think this is not needed

daniel86: maybe it would be good to provide an interface to read settings in C++ code, basically doing a call of `setting/2` in the Pl engine

sasjonge: > but there is a setting to configure this _mng_client:db_name_, can't we use this instead of env?
> maybe it would be good to provide an interface to read settings in C++ code, basically doing a call of setting/2 in the Pl engine

I haven't found an easy way to access the pl settings. I thought that to get them from the pl engine i had to start a PL engine in the already running PL Engine.

> why to append the DB name anyway? I think this is not needed
 
The db name is important for the authentication. If no db is in the string, the user authenticates agains the admin database, which will not work in our case

daniel86: no it should be possible straight forward, I think.
Maybe there is something strange with using C++ API of swipl, I had a strange situation where I thought that the C++ API does not like to do queries using the Pl engines managed by rosprolog. However, I had no issue with using the C API.

daniel86: This is an example for using the C++ API:

```C++
  PlQuery q("call", PlTermv(A2));
  while( q.next_solution() )
  { sum += (long)A1;
    n++;
  }
```

daniel86: here is an example for C API: https://github.com/daniel86/rosprolog/blob/master/src/rosprolog_node/PrologNode.cpp#L61

daniel86 (approved):

PR #264, opened 2020-11-18
Neem validation checks

kaviyachandran:


src/ros/neem_validation.pl
```diff
@@ -0,0 +1,155 @@
...
+  length(Joints, Listlength),
+  ( Listlength > 0 -> print_message(warning, 'The following joints have no parent link or child link asserted. Please assert the corresponding links for the joints'),
+  forall(member(O, Joints), print_message(warning, O))), !;
+  print_message(info, 'All joints have their corresponding links'), true.
+
+load_logs(Folder) :-
+  get_path(Path, Folder),
+  remember(Path),
+  tf_mng_remember(Path).
+
```
sasjonge, line 120: Add a predicate that calls all the test methods for the current episode (so that user can call it at the end of the neem generation)

sasjonge (changes requested):
Generally I think ros is not the correct sub directory. Maybe model would be better

src/ros/neem_validation.pl
kaviyachandran: okay sure. I will update it and move the file

kaviyachandran: Modified to satisfy the requested changes

src/model/neem_validation.pl
```diff
@@ -0,0 +1,143 @@
...
+  get_objects_without_location(ObjectWithoutLocation),
+  get_objects_without_shape(ObjWithoutShape),
+  get_joints_without_proper_links(JointWithoutPLinkOrCLink).
+
+
+
+     /*******************************
+     *          UNIT TESTS          *
+     *******************************/
+:- begin_tests('neem_validation').
```
sasjonge, line 138: I think the current test code does much, therefore I would remove it

sasjonge (changes requested):

src/model/neem_validation.pl
kaviyachandran: Done. Fixed your comments

sasjonge (approved):

PR #265, opened 2020-11-18
Fixing minor issues in obj.pl

kaviyachandran:
- Fixing the options for tell all case in object_dimensions and object_color_rgb
- Add a missing parenthesis in object_shape


daniel86 (approved):
cool!

PR #277, opened 2021-02-01
Fix import of json data

kaviyachandran:
This PR fixes the issue with importing json data

### DIFF
--- a/src/lang/terms/triple.pl
+++ b/src/lang/terms/triple.pl
@@ -72,19 +72,25 @@ triple_import_json(FilePath) :-
 	close(Stream).
 
 read_data(Stream,[]):-
-  at_end_of_stream(Stream).
+	at_end_of_stream(Stream).
 
 read_data(Stream,[TriplesDict | Rest]):-
-  \+ at_end_of_stream(Stream),
-  json:json_read_dict(Stream, TriplesDict),
-  assert_triple_data(TriplesDict),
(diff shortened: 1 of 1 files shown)


src/lang/terms/triple.pl
```diff
@@ -72,19 +72,25 @@
...
 
 assert_triple_data(Triples) :-
-	term_to_atom(Triples.get(s), S),
-	term_to_atom(Triples.get(p), P),
-	term_to_atom(Triples.get(o), O),
-	tell(triple(S, P, O)).
+	get_dict(s, Triples, S),
+	get_dict(p, Triples, P),
+	get_dict(o, Triples, O),
+	(is_dict(O) ->
```
daniel86, line 86: I think a comment would be good why O could be a dictionary here. It seems strange to me

src/lang/terms/triple.pl
```diff
@@ -72,19 +72,25 @@
...
+	findall(Value, (Value = O.Key), Values),
+	member(Object, Values),
+	atom_string(O_atom,Object);
```
daniel86, line 89: It is suggested to not write the *;* at the end of a line to avoid people not seeing it properly.

```
( Goal1
; Goal2
)
```

instead of
```
Goal1;
Goal2.
```

daniel86 (changes requested):

daniel86 (approved):

PR #288, opened 2021-03-31
swrlx:makeOWLIndividual support

Thordreck:
### Summary
Hi,
This pull request aims to give support for the `swrlx:makeOWLIndividual` swrl extension built-in. This extension allows to generate new individuals inside a rule's body. The ontology containing the swrlx built-ins can be found [here](http://swrl.stanford.edu/ontologies/built-ins/3.3/swrlx.owl). A description of the command's behavior is available in [Protege's github](https://github.com/protegeproject/swrlapi/wiki/ExtensionsBuiltInLibrary) (note: that link mentions the `makeOWLThing` command instead, but it's just an alias for `makeOWLIndividual`).

### Use Case

Being able to instantiate individuals inside a rule's body can be useful to generate new events based on the information already available in the knowledge base. Consider the following rule:

```
Person(?p), associatedWith(?p, ?as), Thing(?as), makeOWLIndividual(?event, ?p, ?as) -> Event(?event).
```

This rule will create a new individual if a person is associated with something. New individuals are created for each unique combination of additional arguments passed to the `makeOWLIndividual` command (`?p` and `?as` in this case). This guarantees that only one individual will be created even if the rule is queried several times using the same values for `?p` and `?as`.
With this addition, queries like `instance_of(Event, dul:'Event')` would now not only check if the requirements for an event to take place are met, but they would also instantiate it in the knowledge base (triggering also notifications). The alternative to this is to write additional code implementing the logic needed to decide when a new event should be generated.
This approach is similar to those used in [1] and [2].

### Proposed Changes

Modifications are limited to the `swrl.pl` script:

1. `swrl_builtin_pl` predicates matching calls to `makeOWLIndividual`.
2. Additional predicates to handle arguments patterns and individuals generation (`swrlx_make_individual/2`, etc).

### Issues

So I ran into some issues while implementing this. Probably because I'm not that used to write Prolog and I don't have a full picture of the codebase either. I'm listing them here so we can discuss better ways to do this.

1. I wanted to add the `swrlx` built-in as a separate module in a different file. Problem is that the script `SWRL/parser.pl` calls the `swrl_builtin_pl` predicate in the `swrl` module to parse all the built-ins. I was not sure how to handle this, so I ended up just dumping everything inside the `swrl` module.
2. As far as I know, the `makeOWLIndividual` command can have any number of arguments. Arguments starting from the second position are used to generate an unique pattern to decide if an individual should be created or not. I did not find a way to pass a list to the `swrl_builtin_pl` predicates, since arguments are expected to a be a list of comma-separated values enclosed between parenthesis. I ended up writing the same predicates with different arities (up to 5 arguments).
3. Asserting the new individual into the knowledge base with the command `tell(instance_of(A, owl:'Thing'))` triggers the `notify` predicate. This predicate checks if the new individual is an event. Now the problem (I think) is that those predicates are tabled, so when the individual is first created it will fail. This makes it so that even if later the rule's head infers that the new thing created is an event (like in the example above), no notifications are triggered. As a workaround I'm inserting it using `triple/3` directly.

### References

[1] Zhu, Tao, et al. "Applying semantic web techniques to reservoir engineering: Challenges and experiences from event modeling." 2010 Seventh International Conference on Information Technology: New Generations. IEEE, 2010.

[2] Alaya, Mahdi Ben, and Thierry Monteil. "FRAMESELF: A generic autonomic framework for self-management of distributed systems-Application on the self-configuration of M2M architecture using semantic and ontology." International Conference on Collaboration Technologies and Infrastructures (IEEE WETICE 2012),. 2012.


daniel86:
@Thordreck interesting! I wasn't aware of this extension, but also had demand for creating new symbols in rules when I tried to apply them for action effects that create new objects.

daniel86:
one testcase is failing for some reason. But I think this particular test will not be relevant anymore once I finally merge my work of last months: https://github.com/daniel86/knowrob/tree/aggregate
might happen this week

src/reasoning/SWRL/swrl.pl
```diff
@@ -301,3 +339,40 @@
 swrl_nums([X|Xs],[Y|Ys],Vars) :-
   swrl_atom_number(X,Y,Vars),
   swrl_nums(Xs,Ys,Vars).
+
+%% swrlx storage
+:- dynamic swrlx_individual_map/2.
+
+swrlx_add_individual(Pattern,Individual) :-
+    (   swrlx_individual_map(Pattern, Individual)
```
daniel86, line 347: I am a bit confused about what this pattern is, and why there is a 1-to-1 mapping between pattern and individual.
Could you add a comment/example?

src/reasoning/SWRL/swrl.pl
```diff
@@ -200,6 +215,29 @@
 swrl_builtin_pl(endsWith(A,X), atom_concat(_, X_atom, A_atom), Vars) :-
   swrl_atoms([A,X], [A_atom,X_atom], Vars).
 
+%% extensions
+% Note: built-in arguments are passed individually. I have not found a way to 
+% pass them as a list. As a workaround I am adding support for up to 5 arguments.
+swrl_builtin_pl(makeOWLIndividual(A, B),
```
daniel86, line 221: Can you add a comment what the arguments are? From the comment it is unclear why built-in arguments are passed at all. Are these al types of the individual?

daniel86 (changes requested):
Could you add some more documentation, and also add some unittest?
Or show some example code in the comment and I can add unittests later.

src/reasoning/SWRL/swrl.pl
daniel86: Maybe you could copy some of the pull request comment into the source code, this would help reading the code

daniel86:
Regarding Issue 1. maybe `swrl_builtin_pl/3` can be defined as multifile predicate? then extension could define additional clauses

daniel86:
Regarding Issue 2. I guess another clause of `swrl_builtin_arg/1` in parser.pl needs to be written to handle lists as arguments of built-ins. Maybe something like this would work:

    swrl_builtin_arg(Args) --> ['['], swrl_builtin_args(Args), [']'].

src/reasoning/SWRL/swrl.pl
```diff
@@ -301,3 +339,40 @@
...
+swrlx_get_individual(Pattern, Individual) :-
+   swrlx_individual_map(Pattern, Individual).
+
+%% swrlx commands
+swrlx_make_individual(A, _) :-
+  ground(A).
+
+swrlx_make_individual(A, PatternArgs) :-
+  \+ ground(A),
+  maplist(ground, PatternArgs),
```
daniel86, line 361: `ground(PatternArgs)` works too!

src/reasoning/SWRL/swrl.pl
daniel86: Shouldn't be some kind of rule id be included in the pattern? In case multiple rules would have the same pattern

daniel86:
Hello @Thordreck. I started integrating this pull request into my fork. There is one aspect that makes it more difficult now: the SWRL rules are implemented such that they can be processed by server-side by mongo DB. This won't work as is for the `swrlx_individual_map/2` facts. But it should be possible to rewrite it a bit so that mongo can handle it, and to move the facts into the mongo DB instead of having them in the Prolog KB. A nice side-effect is that the map is persisted then, and you can restart KnowRob without creating redundancies in the triple store.
Also the scope is not handled at all in my fork at the moment, and all inferred facts are assumed to be universally true, so I just left a FIXME note for the moment to allow scoping again in SWRL rules (shouldn't be hard to add again).

daniel86:
I was able to merge this request into my fork. It will be merged into master soon

daniel86:
The issues should now also be resolved. Here is the code: https://github.com/daniel86/knowrob/blob/master/src/reasoning/SWRL/swrlx.pl

daniel86:
@Thordreck usually there is a authors list in the comment at the top, I could list you there if you like

Thordreck:
Hi Daniel, thanks a lot for your work on this pull request. I did not have time to look into this recently. Sorry for not answering your requests for changes and questions. I checked the code in [your personal fork](https://github.com/daniel86/knowrob/blob/master/src/reasoning/SWRL/swrlx.pl) and it's looking great.
Judging from the comments it seems you already figured everything out on your own, but I'll try to answer your questions anyway:

> So I guess these arguments are there to create a pattern for an individual. A type of signature of the individual within the SWRL rule to avoid that it is instantiated multiple times?

Yes, a unique pattern of individuals generates a single individual once. Subsequent queries using the same pattern simply retrieve that newly-added individual.

> So basically the person writing the rule then decides for which different groundings in a query a new individual shall be generated?

That's it. Whoever writes the rule gets to decide what combination of individuals are to be considered to generate said new individual. As I mentioned in the pull-request's description, I'm using this pattern to generate events based on changes in the KB. In this context, I consider this individual pattern as some kind of "event context". For example, consider the following rule that generates a new event when a new person appears:

```
Person(?p), makeOWLIndividual(?event, ?p) -> PersonAppearsEvent(?event).
```

And that's it, for each unique person's individual I know a single event will be generated.

> Shouldn't be some kind of rule id be included in the pattern? In case multiple rules would have the same pattern

Yep, as it's right now it's possible to have individual pattern clashing if different rules happen to use the same combination of individuals. I didn't really think of this, to be honest. I think the original Protege's implementation does not take this into account, either. I might look into this later down the road. Thanks for pointing it out.

> the SWRL rules are implemented such that they can be processed by server-side by mongo DB. This won't work as is for the swrlx_individual_map/2 facts. But it should be possible to rewrite it a bit so that mongo can handle it, and to move the facts into the mongo DB instead of having them in the Prolog KB

Looking at the code in your fork I think you have already done this, haven't you? If not, let me know and I could take a look into it.

> Also the scope is not handled at all in my fork at the moment, and all inferred facts are assumed to be universally true, so I just left a FIXME note for the moment to allow scoping again in SWRL rules (shouldn't be hard to add again).

Same as above, if there's anything I can do to help I would be happy to do so.

> usually there is a authors list in the comment at the top, We could list you there if you like

That would be great! Thanks a lot.

daniel86:
> Yep, as it's right now it's possible to have individual pattern clashing if different rules happen to use the same combination of individuals. I didn't really think of this, to be honest. I think the original Protege's implementation does not take this into account, either. I might look into this later down the road. Thanks for pointing it out.

I figured something out. Basically using the "label" of the rule as first element in the pattern. If there is no human readable label given, then a hash of the rule is used instead.

daniel86:
> Looking at the code in your fork I think you have already done this, haven't you? If not, let me know and I could take a look into it.

yes, it was easy to move the facts into the mongo DB, I did rename the predicate because it will pop up as a collection name in mongo db. So I thought better to include a "swrlx" prefix there.

daniel86:
> That would be great! Thanks a lot.

Can you provide me your name and mail address that should be listed?

Thordreck:
> Can you provide me your name and mail address that should be listed? Or you could add them in a pull request if you like

Sure, I'll just add it here.

Name: Álvaro Páez
Email: paezguerraalvaro@gmail.com

daniel86:
> Same as above, if there's anything I can do to help I would be happy to do so.

It is mostly a matter of adding scope argument to kb_call and  to pass it on to kb_project (https://github.com/knowrob/knowrob/blob/master/src/reasoning/SWRL/swrl.pl#L63).
But kb_call is called in a findall, and then kb_project only once with the list of results.
Expecting that results may have different scopes, it would then be needed to call kb_project once for each different scope, or doing one additional call of kb_call and wrapping each scoped fact in during\2 and assert/1 so that you have a list like `[during(assert(triple(_,_,_)), Scope), ...]`. Not sure if it would work the other way, i.e., `assert(during(triple(_,_,_),Scope))`, but it would look more intuitive for sure.

PR #291, opened 2021-04-13
Update README.md

SDBonhof:
Add `wstool init` to installation instructions in case a wstool workspace hasn't been created yet.

This line can save a lot of time for people who aren't as familiar with wstool. If their workspace already has a wstool workspace then this extra line does no harm.


daniel86:
thanks a lot @SDBonhof !

daniel86 (approved):

Issue #305, opened 2021-09-27
How to access mongo without access control?


JeremiasThun:
The KnowRob MongoInterface knows two ways to communicate with mongo: either via localhost with no specifications at all or with authentication and everything. But what if I want to connect to a MongoDB with a host other than localhost and no access control (as is the default for mongo installations)? 

It doesn't seem to be possible right now.

https://github.com/knowrob/knowrob/blob/4d7af5484683cf2ff0aedd3b0126ae8c5077e035/src/db/mongo/MongoInterface.cpp#L97-L109

JeremiasThun:
In my case, this other mongo host is a docker container, running in a GitLab CI/CD pipeline as Service.

daniel86:
The handling is better now, user and password are optional, see https://github.com/knowrob/knowrob/blob/dev/src/storage/mongo/MongoKnowledgeGraph.cpp#L173

PR #306, opened 2021-09-27
Fix remote mongo login without authentication

JeremiasThun:
As stated in Issue #305, KnowRob doesn't allow a connection to Mongo that is remote and without authentication, right now. 

When these two environment variables are set like this
```
KNOWROB_MONGO_HOST=localhost # or any remote. The error stays the same
KNOWROB_MONGO_PORT=27017
```
KnowRob shuts down with the following error message:
```
ERROR:    Unhandled exception: mng_error(invalid_uri('Invalid host string in URI'))
[ERROR] [1632778849.817918826]: mng_error(invalid_uri('Invalid host string in URI'))
[ERROR] [1632778849.818084663]: Failed to load initial_package knowrob.
[ERROR] [1632778849.818247573]: rosprolog service failed to initialize.
```

Since there are (possibly rare) use cases where users need such a setup, I propose this fix.
In my case, I needed this as part of a dind setup.

### DIFF
--- a/src/db/mongo/MongoInterface.cpp
+++ b/src/db/mongo/MongoInterface.cpp
@@ -8,7 +8,6 @@
 
 #include <mongoc.h>
 #include <sstream>
-#include <iostream>
 
 #include <rosprolog/rosprolog_kb/rosprolog_kb.h>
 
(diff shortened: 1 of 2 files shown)


sasjonge:
Hello Jeremias,

if I remember right this fix made problems and you used my KnowRob Image in the end. Could we therefore close this PR?

Best, Sascha

PR #307, opened 2021-09-28
tf-mongo bugfixes

benjaminalt:
Two small MongoDB-related fixes, one for collection_name in lang/db.pl, one for tf_mng_tree in tf_mongo.pl.

### DIFF
--- a/src/ros/tf/tf_mongo.pl
+++ b/src/ros/tf/tf_mongo.pl
@@ -114,7 +114,7 @@ tf_mng_lookup(ObjFrame, PoseData) :-
 
 %% tf_mng_trajectory(+Obj, +Begin, +End, -Trajectory) is det.
 %
-% Read all transforms associated to a frame in given time interval. 
+% Read all transforms associated to a frame in given time interval.
 %
 tf_mng_trajectory(Obj,Stamp0,Stamp1,Trajectory) :-
 	rdf_split_url(_,ObjFrame,Obj),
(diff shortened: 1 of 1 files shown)


src/lang/db.pl
```diff
@@ -38,15 +38,15 @@
...
-:- multifile collection_name/1.
-
 %%
 annotation_property('http://www.w3.org/2000/01/rdf-schema#comment').
 annotation_property('http://www.w3.org/2000/01/rdf-schema#seeAlso').
 annotation_property('http://www.w3.org/2000/01/rdf-schema#label').
 annotation_property('http://www.w3.org/2002/07/owl#versionInfo').
 
+collection_name(Name) :-
+	mng_db_name(DB), mng_collections(DB, Collections), !, member(Name, Collections).
```
sasjonge, line 48: This fix is problematic, because you don't want to always import or export all collections you have. The ones that are necessary are most of the time only triples and tf. 

At the moment the user could choose the collections by asserting collection_name themselves. I think i could either add a setting in which the user can have a list of the collections he wants to ex- or import (default would be triple and tf), or one could just fix this two collections here and allow the user to assert additional collections.

src/ros/tf/tf_mongo.pl
```diff
@@ -226,11 +226,11 @@
 
 tf_mng_tree(Stamp, Tree) :-
 	% read all required transforms
-	lookup_transforms(Stamp, Transforms),
+	tf_mng_lookup_all(Transforms, Stamp),
```
sasjonge, line 229: I tested tf_mng_tree with your changes and it works. Therefore I would merge atleast this part.

sasjonge (changes requested):

src/ros/tf/tf_mongo.pl
benjaminalt: Sounds good!

src/lang/db.pl
benjaminalt: Sounds good. I removed it from this PR, but I think that this should still be somehow addressed. Currently, the default behavior of remember/1 is to not import anything, because collection_name is always false. So adding default values/a setting for collection_name would be very useful.

sasjonge (approved):

PR #308, opened 2021-09-28
Support for storing and loading wrenches

benjaminalt:
Add support for storing and loading subsymbolic data about wrenches
(forces & torques) acting on objects. This works exactly like the
mechanism for loading and storing /tf.

Right now, wrench data is always defined in the local coordinate
system of the object they act on and cannot be transformed by KnowRob.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,7 @@ target_link_libraries(mongo_kb
 add_library(tf_knowrob SHARED
 	src/ros/tf/tf.cpp
 	src/ros/tf/memory.cpp
-	src/ros/tf/logger.cpp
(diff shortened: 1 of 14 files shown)


sasjonge:
Hello Benjamin,

this PR seems to me to be very specific for your project? I'm not sure if we should add it to the general KnowRob code base. 

Maybe a KnowRob package with an own repository would be the better choice.

Best, Sascha

benjaminalt:
Hi Sascha,
alright! I just thought that if force-dynamic events (like contacts) are stored symbolically in NEEMs, maybe it would be useful to have access to the subsymbolic force data as well. I'll keep this in my local KnowRob fork for now and maybe spin it off into a separate package later.

PR #317, opened 2021-11-25
Prefix bug

AbhijitVyas:


src/lang/mongolog/mongolog.pl
```diff
@@ -135,7 +135,13 @@
 	% TODO: make the client return docs in a format that it accepts as input.
 	maplist(format_doc, Docs, Docs0),
 	maplist(bulk_operation, Docs0, BulkOperations),
-	mng_get_db(DB, Coll, Key),
+
+	%% check for prefix in the Key
+	once(((setting(mng_client:collection_prefix, Prefix),
```
sasjonge, line 140: Maybe it would be a bit clearer and you wouldn't have to use once if you use implication here

src/ros/tf/tf_mongo.pl
```diff
@@ -40,7 +40,7 @@
    		% time field
    		'header.stamp',
 		% options
-		[	collection(CollectionName)
+		[	collection(tf)
```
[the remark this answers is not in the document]
daniel86: @AbhijitVyas ?

src/lang/mongolog/mongolog.pl
daniel86: @AbhijitVyas ?

AbhijitVyas:
For some unknown reason, all comments I made were in pending state, which means no one except me can see them. Hence I am writing new comment in reply to @sasjonge's comments above.

AbhijitVyas:
> Why is this fix necessary? I think this could be a problem with neems, when there is a prefix

it was due to the fact that the CollectionName somehow was repeating the prefixes we provide while creating the collection name.

AbhijitVyas:
> Maybe it would be a bit clearer and you wouldn't have to use once if you use implication here
It works with `once` and I did not get his comment on first time, but after the verbal conversion we agreed to keep it this way in PR.

AbhijitVyas:
@sasjonge I would request you to consider reviewing one more time and resolve the issues if you are satisfied with my answers here. Incase you have further questions/comments, write them out! :)

AbhijitVyas:
> 

Okay, I would agree if we are doing this way when we specifically make knowrob settings with readOnly mode then there should not be an issue of duplication of prefixes.

PR #326, opened 2022-06-09
Noetic

hawkina:
Hi
I've set rosprolog to pull from the noetic branch instead of master since otherwise it causes errors and the workspace won't compile. In the rosinstall it is set to my fork now of knowrob so that the installation instructions would work. It would be nice if we could have a noetic branch or a noetic rosinstall where the correct version of rosprolog is pulled :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ We have prepared different *.rosinstall* setup files that you can add to your RO
 rosdep update
 cd ~/catkin_ws/src
 wstool init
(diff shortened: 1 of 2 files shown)


hawkina:
> Hey, could you create a new PR to the noetic branch and close this one? Also you have still your namespace hawkina in the urls.

ye sure will do. 
yeah I kinda did that so that it would work on my fork as a hotfix solution. i'll change it :)

PR #351, opened 2022-10-16
Work on URDF handling

daniel86:
- removed hasBaseLinkName and hasEndLinkName properties from urdf.owl. only hasBaseLink and hasEndLink should be used in robot and environment owl files
- also removed predicates `has_base_link_name/2` and `has_end_link_name/2`
- convention: entities in the OWL can be classified as urdf:'Link', their name should in this case appear as a frame name in the associated URDF file
- for shapes it is needed that such urdf:'Link''s appear, and also that the component hierarchy is properly represented (in case individual objects within the urdf are displayed)
- added `urdf_pose` and `urdf_pose_rel` to read pose data from URDF file (used to show an object in base pose in openEASE)
- several smaller things, slightly more efficient database interaction


sasjonge (approved):
Tested it locally and it seems to work

PR #355, opened 2022-10-20
Fix mng_import

sasjonge:
mng_import would return false if one of the collections doesn't exist in the directory to be imported. This fix will have mng_import return true as long as one collection to be imported exist.

I also removed inferred from the default import list, because inferred isn't created automatically anymore.

### DIFF
--- a/src/lang/db.pl
+++ b/src/lang/db.pl
@@ -40,7 +40,7 @@
 		'List of named graphs that should initially by erased.').
 
 %%
-:- setting(mng_client:collection_names, list, [triples, tf, annotations, inferred],
+:- setting(mng_client:collection_names, list, [triples, tf, annotations],
 		'List of collections that will be imported/exported with remember/memorize.').
 
 %%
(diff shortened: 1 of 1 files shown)


src/lang/db.pl
```diff
@@ -69,13 +69,20 @@
...
-			mng_get_db(DB, Collection, Name)
+	findall((DB, Dir0),
+		(	
+			collection_name(Name),
+			mng_get_db(DB, Collection, Name),
+			path_concat(Dir, Collection, Dir0),
+			exists_directory(Dir0)
 		),
-		(	path_concat(Dir, Collection, Dir0),
-			mng_restore(DB, Dir0)
```
daniel86, line 77: Wouldn't it be a bit more elegant to write
```
( exists_directory(Dir0) -> mng_restore(DB, Dir0) ; log_warn(...) )
```

src/lang/db.pl
```diff
@@ -69,13 +69,20 @@
...
-		)
+		DirCollection
+	),
+	% Fails if there is no directory to import
+	not(length(DirCollection, 0)),
```
daniel86, line 82: this should not silently fail, a warning should be logged

daniel86 (changes requested):

daniel86:
Closing due to inactivity

PR #365, opened 2023-01-27
Added ros interface to blackboard implementation

sasjonge:


src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+        numSolutions_ += 1;
+        return !has_stop_request_;
+    }
+
+    bool has_more_solutions() {
+        return numSolutions_ > 0;
+    }
+
+    std::string next_solution() {
+        std::string solution = solutions_.front()->c_str();
```
daniel86, line 88: remove *c_str()* it returns a `char*` and through assignment you convert it to *std::string* again, so the call is effectively not needed

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+// Override QueryResultHandler
+    bool pushQueryResult(const QueryResultPtr &solution) override {
+        if(solution->substitution()->empty()) {
+            solutions_.push_back(std::make_shared<std::string>("True."));
+        }
+        else {
+            std::stringstream ss;
+            ss << *solution->substitution() << ".";
+            solutions_.push_back(std::make_shared<std::string>(ss.str()));
```
daniel86, line 77: are you sure that the format is as expected? In rosprolog, the JSON encoding of the solution string was done in Prolog code, here: https://github.com/knowrob/rosprolog/blob/master/src/query.pl#L36

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+	  currentQuery_(""),
+      hybridQA_(config)
+	{
+        incrementalMode = incremental;
+        errorMessage = "";
+        currentQuery_ = queryString;
+        runQuery(currentQuery_);
+	}
+
+    std::string substitutionToString(const QueryResultPtr &sharedPtr) {
```
daniel86, line 65: maybe it's better to replicate the JSON formatting code here

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+            res.solution = solution;
+        }
+    }
+    return true;
+}
+
+bool finish(json_prolog_msgs::PrologFinish::Request &req,
+            json_prolog_msgs::PrologFinish::Response &res)
+{
+    if (req.id == "*"){
```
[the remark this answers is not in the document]
sasjonge: It is a string

src/ROSInterface.cpp
sasjonge: I rewrote substitutionToJSONString to be used inteast

sasjonge:
@daniel86 I think I now incorporated every point of your comments. 

I rewrote a substitution to string function, that works with the rosprolog_commandline.

Additionally I added the tf code. The code compiles now, but I still have to test it.

sasjonge:
Something weird happened with the diff in this PR for PrologReasoner.h. I simply added an additional version of the eval function.

include/knowrob/ros/tf/logger.h
```diff
@@ -29,6 +29,9 @@ class TFLogger
 	void set_db_name(const std::string &db_name)
 	{ db_name_ = db_name; }
 
+    void set_db_uri(std::string db_uri)
```
daniel86, line 32: const std::string &db_uri

include/knowrob/ros/tf/republisher.h
```diff
@@ -36,6 +36,9 @@ class TFRepublisher
 	void set_db_name(std::string db_name)
 	{ db_name_ = db_name; }
 
+    void set_db_uri(std::string db_uri)
```
daniel86, line 39: const&

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+#define PARAM_INITIAL_GOAL "initial_goal"
+#define PARAM_NUM_ROS_THREADS "num_ros_threads"
+#define PARAM_ENABLE_DEBUG "pl_debug"
+
+#define NUM_ROS_THREADS_DEFAULT 2
+
+using namespace knowrob;
+namespace po = boost::program_options;
+
+class HybridQAResultHandler : public QueryResultHandler {
```
daniel86, line 41: probably should be named differently, *ROSQueryResultHandler* ?

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+    std::list<std::shared_ptr<std::string>> solutions_;
+    HybridQAResultHandler(const boost::property_tree::ptree &config, const std::string &queryString, bool incremental = false)
+	: has_stop_request_(false),
+	  currentQuery_(""),
+      hybridQA_(config)
+	{
+        incrementalMode = incremental;
+        errorMessage = "";
+        currentQuery_ = queryString;
+        runQuery(currentQuery_);
```
daniel86, line 54: it seems strange that the "result handler" itself is issuing a query. it was supposed to only receive a solution. So the class name HybridQAResultHandler seems not appropriate.

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+    std::string substitutionToJSONString(const std::shared_ptr<Substitution> &omega) {
+        auto ss = std::stringstream();
+        uint32_t i=0;
+        ss << "{\n";
+        for(const auto &pair : *omega) {
+            if(i++ > 0) ss << ",\n";
+            ss << '\"' << pair.first.name() << "\":" << '\"' << (*pair.second) << "\"";
```
daniel86, line 63: sometimes pair.second might be quoted with single quotes! probably you need to remove them in case

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+        return ss.str();
+    }
+
+// Override QueryResultHandler
+    bool pushQueryResult(const QueryResultPtr &solution) override {
+        if(solution->substitution()->empty()) {
+            solutions_.push_back(std::make_shared<std::string>("True."));
+        }
+        else {
+            std::stringstream ss;
```
daniel86, line 75: stringstream is not needed, just pass the string as argument

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+            ss << substitutionToJSONString(solution->substitution());
+            solutions_.push_back(std::make_shared<std::string>(ss.str()));
```
daniel86, line 77: probably unneeded to use smart pointer here

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+            auto query = hybridQA_.parseQuery(queryString);
+            // evaluate query in hybrid QA system
+            numSolutions_ = 0;
+            hybridQA_.runQuery(query, *this);
+            if(numSolutions_ == 0) {
+                solutions_.push_back(std::make_shared<std::string>("False."));
+            }
+        }
+        catch (std::exception& e) {
+            std::cout << e.what() << std::endl;
```
daniel86, line 119: you should use logging macros, e.g. `KB_ERROR`

src/ROSInterface.cpp
```diff
@@ -0,0 +1,348 @@
...
+        it = resultHandlers_.begin();
+    }
+}
+
+bool query(json_prolog_msgs::PrologQuery::Request &req,
+           json_prolog_msgs::PrologQuery::Response &res)
+{
+    if (exists(req.id)) {
+        std::stringstream ss;
+        ss << "Another query is already being processed with id " << req.id << ".";
```
daniel86, line 177: the point you can put in single quotes

src/ros/tf/logger.cpp
```diff
@@ -25,13 +25,14 @@ TFLogger::~TFLogger()
...
+    std::shared_ptr<MongoCollection> collection_ = MongoInterface::get().connect(
+            db_uri_.c_str(), db_name_.c_str(),topic_.c_str());
 	if(!mongoc_collection_insert(
-			(*collection)(),MONGOC_INSERT_NONE,doc,NULL,&err))
+			(*collection_)(),MONGOC_INSERT_NONE,doc,NULL,&err))
 	{
 		ROS_WARN("[TFLogger] insert failed: %s.", err.message);
 	}
-	delete collection;
+	collection_.reset();
```
daniel86, line 35: what is the reset doing? is it needed? also for changes below

daniel86: reset is not needed, just remove it everywhere. smart pointer deletes itself, no reset needed.

src/ros/tf/republisher.cpp
```diff
@@ -35,7 +35,7 @@ TFRepublisher::~TFRepublisher()
 	thread_.join();
 	tick_thread_.join();
 	if(collection_) {
-		delete collection_;
+		collection_.reset();
```
daniel86, line 38: reset is not needed, just remove it everywhere. smart pointer deletes itself, no reset needed.

src/ros/tf/republisher.cpp
```diff
@@ -137,11 +137,12 @@ void TFRepublisher::create_cursor(double start_time)
 	if(cursor_!=NULL) {
 		mongoc_cursor_destroy(cursor_);
 	}
-	if(collection_) {
-		delete collection_;
-	}
-	collection_ = MongoInterface::get_collection(
-		db_name_.c_str(),db_collection_.c_str());
+    if(collection_) {
+        collection_.reset();
```
daniel86, line 141: reset is not needed, just remove it everywhere. smart pointer deletes itself, no reset needed.

src/ros/tf/republisher.cpp
```diff
@@ -203,8 +204,8 @@ void TFRepublisher::set_initial_poses(double unix_time)
 	if(append_opts) {
 		bson_destroy(append_opts);
 	}
-	if(collection) {
-		delete collection;
+	if(collection_) {
+		collection_.reset();
```
daniel86, line 208: reset is not needed, just remove it everywhere. smart pointer deletes itself, no reset needed.

daniel86 (changes requested):

src/ROSInterface.cpp
sasjonge: I'm also not completly happy with the name, but i think I want to use the class also for an REST interface later, so ROS doens't fit perfectly. I was thinking about IncrementalQueryResultHandler, but that could create confusion, and make people think the generation of the Results is incremental and not just the next_solution call

src/ROSInterface.cpp
sasjonge: I added a filtering for sourrounding single quotes.

sasjonge:
@daniel86 I incorporated all your comments and splitted my ROSInterface file so that the ResultHandler is in it's own file (for later use with a REST interface). I now named the ResultHandler IterativeQAResultHandler.

daniel86 (approved):

PR #366, opened 2023-01-27
Added noetic dockerfile

sasjonge:


daniel86:
closing this one as we do not really work on master branch any longer.
Would be nice though if we could fix the dockerfiles in dev branch!

PR #372, opened 2023-08-01
Started ROSInterface

sasjonge:
This PR adds an first version of an ros interface for the modal code. It builds and atleast answers simple queries. There are still some todo's like the modalities and some answer types.


sasjonge:
I added modalities and answer parsing. I also added the askone server.

sasjonge:
The changes to the dockerfile are still work in progress, but atleast the base image can be used with a docker toolchain in Clion

include/knowrob/queries/AnswerError.h
```diff
@@ -0,0 +1,45 @@
...
+#include <fmt/core.h>
+#include "knowrob/terms/Term.h"
+#include "Answer.h"
+#include <knowrob/queries/ModalQuery.h>
+
+namespace knowrob {
+    /**
+     * A answer-related runtime error.
+     */
+    class AnswerError : public std::runtime_error {
```
daniel86, line 22: Why don't you use  `QueryError`?

include/knowrob/ros/ROSInterface.h
```diff
@@ -0,0 +1,60 @@
...
+#include "knowrob/knowrob.h"
+#include "knowrob/Logger.h"
+#include "knowrob/KnowledgeBase.h"
+#include "knowrob/queries/QueryParser.h"
+#include "knowrob/formulas/ModalFormula.h"
+// ROS
+#include <ros/ros.h>
+#include <ros/console.h>
+#include <ros/package.h>
+#include <iai_knowledge_msgs/GraphAnswer.h>
```
daniel86, line 24: Why "iai_knowledge_msgs"? Could we use `<knowrob/ros/msgs>` instead? External people might not associate "iai" with knowrob, so I suggest to put knowrob into the name

include/knowrob/ros/ROSInterface.h
```diff
@@ -0,0 +1,60 @@
...
+        virtual ~ROSInterface();
+
+        void executeAskAllCB(const iai_knowledge_msgs::askallGoalConstPtr &goal);
+
+        void executeAskOneCB(const iai_knowledge_msgs::askoneGoalConstPtr &goal);
+
+        // void executeAskIterativeCB(const iai_knowledge_msgs::askiterativeGoalConstPtr &goal);
+
+        static FormulaPtr
+        applyModality(const iai_knowledge_msgs::askallGoal_<std::allocator<void>>::_query_type &query,
```
daniel86, line 53: the first argument looks a bit wild, why is it written different compared to the source file? Shouldn't it be `const iai_knowledge_msgs::GraphQuery &query` ?

src/ros/ROSInterface.cpp
```diff
@@ -0,0 +1,268 @@
...
+    askone_action_server_.start();
+    //askiterative_action_server_.start();
+}
+
+ROSInterface::~ROSInterface() = default;
+
+FormulaPtr
+ROSInterface::applyModality(const iai_knowledge_msgs::GraphQuery &query,
+                            FormulaPtr phi) {
+    FormulaPtr mFormula = std::move(phi);
```
daniel86, line 49: not sure about using std:move on shared pointer here. It might nullify `phi`! Probably better just to create a copy and to keep `phi` alive too. also you create a copy (of the shared pointer) whenever this function is called because you do not pass by reference. So maybe rather use `const FormulaPtr &phi` to avoid copy on calling the function

src/ros/ROSInterface.cpp
```diff
@@ -0,0 +1,268 @@
...
+    askall_action_server_.setSucceeded(result);
+}
+
+void ROSInterface::executeAskOneCB(const iai_knowledge_msgs::askoneGoalConstPtr& goal)
+{
+    // Implement your action here
+    FormulaPtr phi(QueryParser::parse(goal->query.queryString));
+
+    FormulaPtr mPhi = applyModality(goal->query, phi);
+    auto query = std::make_shared<ModalQuery>(mPhi, QUERY_FLAG_ONE_SOLUTION);
```
daniel86, line 205: I think this is doing nothing, you can just pass mPhi to submitQuery

src/ros/ROSInterface.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+    auto resultStream = kb_.submitQuery(query->formula(), QUERY_FLAG_ONE_SOLUTION);
```
daniel86, line 207: I was rather thinking we use the other submitQuery interface here where you provide a `GraphQuery` with a `ModalityFrame`. then we could rather construct the `ModalityFrame` directly from the ROS message. An advantage would be that the graph query interface takes a shortcut compared to if you submit a modal query

src/ros/ROSInterface.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+    return config;
+
+}
+
+int main(int argc, char **argv) {
+	InitKnowledgeBase(argc, argv);
+
+    // Load settings files
+    loadSetting();
```
daniel86, line 258: seems nothing is done with settings here! The `KnowledgeBase` should be initialized with the settings

daniel86 (changes requested):

include/knowrob/queries/AnswerError.h
sasjonge: I thought QueryError is more specific for problems with the query or the querying. AnswerError is an error if the answer is in way that is erroenous. This is only used when parsing the answers for the ros action. If there is a modal operator this error would be thrown.

include/knowrob/ros/ROSInterface.h
sasjonge: I thought this was the standard we used with all software packages of our institute, e.g. like the json_prolog_msgs. But I should add the dependencie atleast to the rosinstall to make it easier to install.

src/ros/ROSInterface.cpp
sasjonge: That call seems to be useless, so I removed it. loadSetting is used to load the knowledgegraph by the call in 261

PR #373, opened 2023-08-21
Add initial_package and initial_goal as args in knowrob.launch

Rezenders:
Add initial_package and initial_goal as args in knowrob.launch


sasjonge (approved):
Seems good!

PR #377, opened 2023-11-23
Negation of statements

daniel86:
This pull request is concerned with support of negated statements in queries. For now, only *closed-world semantics* of negation is considered.

- so a query “~p” can be considered true if every reasoner agrees that “p” is false, and in addition the literal is not stored in EDB in positive form
- evaluation of negated literals should be deferred

note that at a later point open-world semantics should be supported too, and ultimatively it would be best to switch to three-valued predicates (true, false, don't know)


sasjonge (approved):
Tests run through and code looks fine

Issue #378, opened 2023-11-24
Modal branch: AskOne does not terminate


sasjonge:
If trying to run a ask-one query the program doesn't terminate and shows this warning:

> [warning] an exception occurred for graph query ((triple(X, swrl_test:hasNumber, Num) ∧ triple(swrl_test:'Fred', swrl_test:hasSibling, X))): mng_error(cursor_error,unrecognized field 'limit').

I wrote a test case to reproduce this problem on the askOneTest Branch:

https://github.com/knowrob/knowrob/blob/0843af51d1f036f48e1bcf25cd48f6042340d6a8/tests/KnowledgeBaseTest.cpp#L196

daniel86:
fixed with #379

PR #379, opened 2023-11-24
Query evaluation with QUERY_FLAG_ONE_SOLUTION set

daniel86:
This pull request addresses the issue #378 

There were actually two problems:
- an exception was thrown indicating that the generated bson document has invalid syntax
- the query did not return in this case

The first problem seems related to a problem related to using `mongoc_cursor_set_limit`. For me seems like a bug in the mongo C driver. Anyways mongoc_cursor_set_limit can be avoided by using `$limit` aggregation stage.

The second problem was a bug on KnowRob's side that communication channels were not terminated by publishing an EOS message.


sasjonge (approved):
Tested and looks good.

PR #385, opened 2024-02-11
Added base exception class and stacktrace printing

sasjonge:
I added a base exception class and some stacktrace printing.


include/knowrob/BaseError.h
```diff
@@ -0,0 +1,75 @@
...
+
+		/**
+		 * Constructor for error with stacktrace.
+		 * @tparam Args fmt-printable arguments.
+		 * @param fmt A fmt string pattern.
+		 * @param stacktrace String representing the captured stack trace.
+		 * @param args list of arguments used to instantiate the pattern.
+		 */
+		template<typename ... Args>
+		explicit BaseError(const char *fmt, const std::string& stacktrace, Args&& ... args)
```
daniel86, line 45: use `std::string_view` instead of `std::string&` to avoid unneeded copy if you use `char*` as argument

include/knowrob/BaseError.h
```diff
@@ -0,0 +1,75 @@
...
+		 * Overrides the what() function to include the stack trace with the error message.
+		 * @return A C-string with the error message and stack trace.
+		 */
+		const char* what() const noexcept override {
+			static std::string fullMessage; // To ensure the lifetime of the returned c_str()
+			fullMessage = std::runtime_error::what(); // Original error message
+			if (!stacktrace.empty()) {
+				fullMessage += "\nStack Trace:\n" + stacktrace;
+			}
+			return fullMessage.c_str();
```
daniel86, line 69: are you sure this is safe? because fullMessage is destroyed when you return, but you return the char* held by fullMessage so I wouldn't be sure that it is still valid after return

include/knowrob/py/wrapper.h
```diff
@@ -25,7 +26,7 @@ namespace knowrob::py {
 			return python::call_method<R>(self, method, python::object(args)...);
 		} catch (const python::error_already_set&) {
 			knowrob::py::handlePythonError();
-			throw ReasonerError("Python error");
+			throw ReasonerError("Python error", boost::stacktrace::stacktrace());
```
daniel86, line 29: the `handlePythonError` does throw an error already or? So this should never be reached I think.

include/knowrob/BaseError.h
mpomarlan: It should be safe:
https://www.learncpp.com/cpp-tutorial/static-local-variables/

Since fullMessage is declared static, it survives as long as the program runs.

include/knowrob/BaseError.h
daniel86, line 45: it is a bit odd that the fmt format string is not directly next to the args. Because usually you write something like `"foo {}", val` and would assume val gets inserted into the format.

It's not possible to have it after `args`?

include/knowrob/BaseError.h
daniel86: ah you are right, I missed that. But that raises another issue, why is it static? Seems like there are multithreading issues that can be caused by this. i.e. if what() is called concurrently one exception could overwrite the message of the other, also possibly causing a segmentation fault

daniel86:
Closing due to inactivity

PR #389, opened 2024-04-11
bugfix: uninitialized variable in PropertyTree

mmrrqq:
Hi there!

While following along the getting started guide the `knowrob-terminal` would not fully start up. I managed to track down the issue to the `delimiter_` variable of `PropertyTree` not being initialized in all constructors. This let to an infinite loop in `PropertyTree::createKeyTerm`.

This PR fixes the issue above. Hope this is useful for you!


daniel86:
Hi Markus, thanks for the bugfix! Probably we should add a compilation flag such that the compiler reports these warnings. Such warnings are displayed in my IDE though, but it is sometimes a bit slow updating the warning annotations so sometimes such warnings remain unnoticed.

PR #398, opened 2024-05-20
Restructured the Readme and added the ROS client lib

marcomasa:
- Adding a client lib that implements a ROS interface using the actions provided by knowrob
- Modified the Cmakelist to include optimization (-O2) and building of the client lib
- Restrucured the readme to be more user-friendly


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.5.0)
-set(CMAKE_CXX_FLAGS "-Wswitch -std=c++0x -pthread ${CMAKE_CXX_FLAGS} -Wno-undef")
+set(CMAKE_CXX_FLAGS "-Wswitch -std=c++0x -O2 -pthread ${CMAKE_CXX_FLAGS} -Wno-undef")
```
daniel86, line 2: I had a strange issue when adding this flag. Maybe it was caused by a bug that is already fixed, still better to check this carefully. I will need to do some local tests with the flag again

CMakeLists.txt
```diff
@@ -79,6 +79,10 @@ add_definitions(-DKNOWROB_BUILD_DIR=\"${CMAKE_BINARY_DIR}\")
 add_definitions(-DBOOST_ALLOW_DEPRECATED_HEADERS)
 add_definitions(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
 
+<<<<<<< HEAD
```
daniel86, line 82: you did not resolve the merge conflict here

CMakeLists.txt
```diff
@@ -92,32 +96,34 @@ if (WITH_ROS1_INTERFACE)
...
+                AskIncrementalNextSolution.action
+                Tell.action
+        )
+
+        generate_messages(DEPENDENCIES
+                        std_msgs
+                        actionlib_msgs)
 
     catkin_package(
+        INCLUDE_DIRS include
```
daniel86, line 125: I think Sascha should review the part regarding ROS client, as he works on the ROS interface

README.md
```diff
@@ -58,22 +91,134 @@ make
...
+ ```bash
+# (once initially) create a directory for the mongoDB data
+sudo mkdir -p ~/data/db
 
-### Development
+# launch mongoDB server with the created path as argument
+sudo mongod --dbpath ~/data/db
+ ```
 
-Any IDE with proper CMake and C++ language support should be able to load the project.
```
daniel86, line 72: where did this go?

README.md
```diff
@@ -113,111 +258,33 @@ An example configuration file is provided in `settings/default.json`:
 For more information about storage backends, please refer to the [Backends](src/storage/README.md) documentation,
 and for more information about reasoners, please refer to the [Reasoner](src/reasoner/README.md) documentation.
 
-### Launching
-
-Being a shared library, KnowRob cannot be launched directly but only in the context
```
daniel86, line 118: Where did this go?

daniel86 (changes requested):
- There is one error in the pukll request where a git conflict was not merged correctly.
- It appears that some text from the README was removed, was this intentional? Seems to me the removed parts should not be removed
- I think @sasjonge  should review everything related to the ROS client, as I am not entirely aware of its current state
- finally I tried some month ago adding -O2, but it for some strange reason was creating problems back then. So I need to check again if it still happens.

review-wise the easiest would be to have at least 3 pull requests: readme, ros client, and compiler flags. But we can also handle all together, just might take a bit more time....

README.md
```diff
@@ -113,111 +258,33 @@ An example configuration file is provided in `settings/default.json`:
...
-### Querying
-
-The core of KnowRob is a querying interface that is built around a custom querying language.
-Its syntax is similar to Prolog, but it is more simplified and not Turing-complete like Prolog is.
-
-For more information on querying in KnowRob, please have a look
-[here](src/queries/README.md).
-
-### Ontologies
+### Using the ROS interface
```
sasjonge, line 268: I think this subsection should link to the subpage for the ros interface. You could then also have most of the client description there.

README.md
```diff
@@ -113,111 +258,33 @@ An example configuration file is provided in `settings/default.json`:
...
-Ontologies are organized in a hierarchy where each ontology is a specialization of another ontology.
-A common distinction is made between foundational (or top-level) ontologies, domain ontologies, 
-and application ontologies.
-A foundational ontology fixes the basic concepts and relationships that are used
-across different domains.
-In KnowRob, we define a domain ontology for the robotics domain, and align it with a
-common foundational ontology.
-Applications can then import the domain ontology and extend it with application-specific concepts
-to cover the specific requirements of the application.
+Once knowrob is launched, it provides four actions to ask queries:
```
sasjonge, line 272: There are now 5 interfaces with the askIncrementalNextSolution part

CMakeLists.txt
marcomasa: okay, we could leave it out if it causes issues. 
i just thought it normally should be in..

CMakeLists.txt
marcomasa: sorry mb, vs code showed that weird

README.md
marcomasa: Do you really need such a comment?

README.md
marcomasa: It can be found under "starting" knowrob.
I renamed that as i think the term "launch" is mostly used in ros

src/ros_client/client_node.cpp
sasjonge, line 1: Maybe move this to src/integration/ros1?

CMakeLists.txt
```diff
@@ -418,8 +424,47 @@ if (WITH_ROS1_INTERFACE AND catkin_FOUND AND CATKIN_DEVEL_PREFIX)
...
+
+	install(
+  		TARGETS knowrob-ros-cli
+  		LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+	)
+
+	## Test node for the client library
+	add_executable(
+		client-node
+		client-node
```
sasjonge, line 456: This doesn't seem to build for me, i think the path to the cpp is missing?

sasjonge (changes requested):
Hey, 

for the ROS part I would request some minor changes, especially moving the client code to integration/ros1. Also some of the documentation could be moved to the ROS Documentation part (see my comments)

Best, Sascha

src/ros_client/client_node.cpp
marcomasa: can do

CMakeLists.txt
marcomasa: i will investigate when i move the client node

CMakeLists.txt
daniel86: yes we can leave it out for this request. But I will try to make tests with the flag set still this week...

daniel86:
The ROS integration was moved into a repository: https://github.com/knowrob/knowrob_ros
Please have a look there if you would like to add a client library.

I hope the docu is now a bit more helpful! (-O2 also added, was quite an improvement!)

PR #401, opened 2024-05-23
Fix to AskIncremental ROS Services

sasjonge:
The AskIncremental ROS Action Services had a bug, where only the last result was the result, stemming from a missing break command. This is fixed by removing the loop. 

I also added a removal of finished ids for queries and added a service to manually finish the id to allow memory management by the user.

I tested the implementation locally, especially also the case when an queryID is "finished" (either because no answers are left, or because a user did manually set it to finish).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,12 @@ if (WITH_ROS1_INTERFACE AND catkin_FOUND)
             Tell.action
     )
 
+    add_service_files(
+            DIRECTORY ros1/srv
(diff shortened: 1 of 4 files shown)


daniel86 (changes requested):
I think you forgot to include the service files. Please have a look if CI goes through before requesting a review

sasjonge:
Hey, sorry, I fixed it now.

daniel86 (approved):

PR #402, opened 2024-06-13
Created test case for using Knowrob as Python Library and corresponding fixes

sasjonge:
I added a test case for using knowrob as a python library for querying. This PR mainly adds the case so that other devs can work with. It also contains some minor fixes and additions that were necessary for the first steps of the unit test case.

### DIFF
--- a/include/knowrob/knowrob.h
+++ b/include/knowrob/knowrob.h
@@ -22,6 +22,11 @@ namespace knowrob {
 	void InitKnowledgeBase(int argc, char **argv);
 
 	/**
+ 	 * Static initialization of the knowledge base without executable name.
+ 	*/
+	void InitKnowledgeBase();
+
+	/**
 	 * Shutdown the knowledge base.
 	 * This will ensure that all worker threads in the global DefaultThreadPool
 	 * are joined.
--- a/include/knowrob/knowrob.h
+++ b/include/knowrob/knowrob.h
@@ -22,11 +22,6 @@ namespace knowrob {
 	void InitKnowledgeBase(int argc, char **argv);
 
 	/**
- 	 * Static initialization of the knowledge base without executable name.
- 	*/
-	void InitKnowledgeBase();
-
-	/**
 	 * Shutdown the knowledge base.
 	 * This will ensure that all worker threads in the global DefaultThreadPool
 	 * are joined.
--- a/src/integration/python/knowrob_module.cpp
+++ b/src/integration/python/knowrob_module.cpp
@@ -78,6 +78,24 @@ static inline void register_triple_types() {
 	boost::python::class_<TripleList>("TripleList").def(boost::python::vector_indexing_suite<TripleList, true>());
 }
 
+static void InitKnowledgeBaseWrapper(boost::python::list py_argv) {
+	int argc = boost::python::len(py_argv);
+	std::vector<std::string> arg_strings;
+	std::vector<char*> argv;
+
+	for (int i = 0; i < argc; ++i) {
+		std::string arg = boost::python::extract<std::string>(py_argv[i]);
+		arg_strings.push_back(arg);
+	}
+
+	for (auto& str : arg_strings) {
+		argv.push_back(&str[0]);
+	}
+
+	// Call the actual InitKnowledgeBase function with the converted arguments
+	knowrob::InitKnowledgeBase(argc, argv.data());
+}
+
 BOOST_PYTHON_MODULE (MODULENAME) {
 	using namespace boost::python;
 	using namespace knowrob::py;
@@ -115,7 +133,6 @@ BOOST_PYTHON_MODULE (MODULENAME) {
 
 	/////////////////////////////////////////////////////
 	// mappings for static functions
-	def("InitKnowledgeBase", static_cast<void(*)()>(&knowrob::InitKnowledgeBase),
-		"Initialize the Knowledge Base.");
+	def("InitKnowledgeBase", &InitKnowledgeBaseWrapper, "Initialize the Knowledge Base with arguments.");
 
 }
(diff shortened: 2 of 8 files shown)


include/knowrob/knowrob.h
```diff
@@ -21,6 +21,11 @@ namespace knowrob {
 	 */
 	void InitKnowledgeBase(int argc, char **argv);
 
+	/**
+ 	 * Static initialization of the knowledge base without executable name.
+ 	*/
+	void InitKnowledgeBase();
```
daniel86: I do not think this is safe! Prolog requires argc and argv being correctly initialized to the executable that runs knowrob, here https://github.com/knowrob/knowrob/blob/dev/src/integration/prolog/PrologEngine.cpp#L79

In my experience not setting this will break Prolog

tests/py/test_boost_python.py
```diff
@@ -66,3 +66,32 @@ def set_xsd_optional(optional):
 	assert optional == XSDType.STRING, "optional is not STRING"
 	optional = XSDType.DOUBLE
 	assert optional == XSDType.DOUBLE, "optional is not DOUBLE"
+
+
+def query_knowledge_base(optional):
+	# Initialize the knowledge base
+	InitKnowledgeBase()
```
daniel86: should not be done in the test. for unit tests it is done here **https://github.com/knowrob/knowrob/blob/dev/tests/gtests.cpp#L6**

daniel86 (changes requested):

tests/py/test_boost_python.py
sasjonge: My thought process here was to also test the init call directly in Python, because this would be how users would use it if they just want to use the shared library in python. It seems to work, but I could check if a second call could break something.

src/integration/python/knowrob_module.cpp
```diff
@@ -76,6 +78,24 @@ static inline void register_triple_types() {
 	boost::python::class_<TripleList>("TripleList").def(boost::python::vector_indexing_suite<TripleList, true>());
 }
 
+static void InitKnowledgeBaseWrapper(boost::python::list py_argv) {
+	int argc = boost::python::len(py_argv);
+	std::vector<std::string> arg_strings;
+	std::vector<char*> argv;
```
daniel86: I think you need to add `static` keyword to ensure lifetime beyond this call

src/integration/python/knowrob_module.cpp
```diff
@@ -76,6 +78,24 @@ static inline void register_triple_types() {
...
+
+	for (int i = 0; i < argc; ++i) {
+		std::string arg = boost::python::extract<std::string>(py_argv[i]);
+		arg_strings.push_back(arg);
+	}
+
+	for (auto& str : arg_strings) {
+		argv.push_back(&str[0]);
```
daniel86: that's not right. use `str[0].data()`

daniel86 (changes requested):

include/knowrob/knowrob.h
sasjonge: I added some wrapper functions for this, that also allow passing the args

tests/py/test_boost_python.py
sasjonge: I did comment this call out for now

src/integration/python/knowrob_module.cpp
```diff
@@ -76,6 +78,33 @@ static inline void register_triple_types() {
 	boost::python::class_<TripleList>("TripleList").def(boost::python::vector_indexing_suite<TripleList, true>());
 }
 
+static void InitKnowledgeBaseWrapper(boost::python::list py_argv) {
+	static int argc = boost::python::len(py_argv);
+	static std::vector<char *> argv;
+
+	for (int i = 0; i < argc; ++i) {
+		std::string arg = boost::python::extract<std::string>(py_argv[i]);
```
daniel86, line 94: here you copy into a new string `arg` with local allocation within the for loop. so it will be invalid after leaving the current call within the loop

src/integration/python/knowrob_module.cpp
```diff
@@ -76,6 +78,33 @@ static inline void register_triple_types() {
...
+		argv.push_back(arg.data());
```
daniel86: here you take the pointer to the local string data. so after leaving the for loop the data in the array will eventually be invalid

daniel86 (changes requested):

src/integration/python/knowrob_module.cpp
```diff
@@ -76,6 +78,38 @@ static inline void register_triple_types() {
 	boost::python::class_<TripleList>("TripleList").def(boost::python::vector_indexing_suite<TripleList, true>());
 }
 
+static void InitKnowledgeBaseWrapper(boost::python::list py_argv) {
+	static int argc = boost::python::len(py_argv);
+	std::vector<std::string> arg_strings;
```
daniel86: as explained on zulip, this must also be static.
The reason, the vector<char*> points to the memory allocated by vector<std::string>. **so both must be static**

daniel86 (changes requested):

src/integration/python/knowrob_module.cpp
```diff
@@ -76,6 +78,38 @@ static inline void register_triple_types() {
 	boost::python::class_<TripleList>("TripleList").def(boost::python::vector_indexing_suite<TripleList, true>());
 }
 
+static void InitKnowledgeBaseWrapper(boost::python::list py_argv) {
```
daniel86, line 83: Calling this function should be protected to avoid that it is called two time.

e.g. add a static boolean flag, initialize it to false, throw an error when it's true when entering the function, and set it to true withing the function

daniel86 (changes requested):

daniel86 (approved):

PR #403, opened 2024-07-25
Prolog example based on LPN

daniel86:
This pull request creates a test case for Prolog which is based in the learn Prolog now tutorial.
For now there is only one first example, additional ones could be added later on.

The pull request also ioncludes a few fixes I made on the way, and removes some of the unused OWL files.


sasjonge (approved):
Code looks good. I also tested it on my local setup (also within my ros workspace).

PR #405, opened 2024-07-30
Added test cases for working with KnowledgeBase in Python

sasjonge:
This PR adds 4 tests cases to test common use cases for using KnowledgeBase in Python.


include/knowrob/terms/Atomic.h
```diff
@@ -57,6 +58,17 @@ namespace knowrob {
 		 */
 		virtual std::string_view stringForm() const = 0;
 
+		/**
+		 * Get the lexical form of this atomic term.
+		 * @return the lexical form of this atomic term.
+		 */
+		std::string humanReadableForm() const {
```
daniel86, line 65: Any reason you did not add this to `Term` instead? IMO it should be a method of Term

src/integration/InterfaceUtils.cpp
```diff
@@ -19,6 +19,19 @@
 
 using namespace knowrob;
 
+// enum for epistemicOperator
+enum EpistemicOperator {
```
daniel86, line 23: It might be confusing that we also have the enum `ModalType` in `ModalOperator.h`.
Can't we unify these?

src/integration/InterfaceUtils.cpp
```diff
@@ -80,7 +93,7 @@ InterfaceUtils::applyModality(const std::unordered_map<std::string, boost::any>
 
 	// Retrieve epistemicOperator and check if it is "BELIEF"
 	auto epistemicOperator = boost::any_cast<int>(options.at("epistemicOperator"));
-	if (epistemicOperator == 1) {
+	if (epistemicOperator == EpistemicOperator::BELIEF) {
```
daniel86, line 96: I am a bit confused about the whole "applyModalOperator" thing. Maybe you can explain to me offline?

src/integration/python/knowrob_module.cpp
```diff
@@ -156,5 +194,6 @@ BOOST_PYTHON_MODULE (MODULENAME) {
 	// mappings for static functions
 	def("InitKnowledgeBaseWithArgs", &InitKnowledgeBaseWrapper, "Initialize the Knowledge Base with arguments.");
 	def("InitKnowledgeBase", &InitKnowledgeBaseFromSysArgv, "Initialize the Knowledge Base using sys.argv.");
+	def("applyModality", &applyModalityWrapper, "Apply a modality to a formula.");
```
daniel86: Why is this here and not in InterfaceUtils.cpp?

src/terms/Atomic.cpp
```diff
@@ -85,6 +85,9 @@ namespace knowrob::py {
 		class_<Atomic, std::shared_ptr<AtomicWrap>, bases<Term>, boost::noncopyable>
 				("Atomic", no_init)
 				.def("stringForm", pure_virtual(&Atomic::stringForm))
+				.def("__repr__", &Atomic::stringForm)
```
daniel86, line 90: better move into Term class? also below

src/terms/Term.cpp
```diff
@@ -109,5 +109,7 @@ namespace knowrob::py {
 				.def("isBlank", &Term::isBlank)
 				.def("isGround", &Term::isGround)
 				.def("variables", pure_virtual(&Term::variables), return_value_policy<copy_const_reference>());
+		// register shared_ptr to Term
+		register_ptr_to_python< std::shared_ptr< Term > >();
```
daniel86, line 114: I think this is not needed. the pointer is registered in the `class_` statement or not?

tests/py/test_boost_python.py
```diff
@@ -5,6 +5,30 @@
...
+		"temporalOperator": TemporalOperator.CURRENTLY,
+		"minPastTimestamp": -1.0,
+		"maxPastTimestamp": -1.0,
+	}
+	# Load the settings
+	kb = KnowledgeBase(settings_path)
+	# Create a formula for the query
+	phi = QueryParser.parse(query_string)
+	# Apply the modality
+	mPhi = applyModality(modalities, phi)
```
daniel86: Why don't you put the modal operator into the query string instead? Then you do not need to apply a modal operator after parsing.

Also there is already a syntax available for doing this. You can find an example here: https://github.com/knowrob/knowrob/blob/dev/tests/KnowledgeBaseTest.cpp#L273

daniel86 (changes requested):

src/integration/InterfaceUtils.cpp
sasjonge: As discussed: It would be nicer for the ROS messages (and also future REST definitions etc.) to have each modaltype start counting with 0

src/integration/InterfaceUtils.cpp
sasjonge: As discussed: This is necessary for ROS or other interfaces that provide modalities in external map like structures

src/terms/Term.cpp
sasjonge: I tested it again without this line, but then the kb_positive_query test fails in line 136 when getting the substitution.

src/knowrob.cpp
```diff
@@ -78,11 +80,61 @@ namespace knowrob {
...
+
+namespace knowrob::py {
+	void staticKnowRobModuleInit() {
+		using namespace boost::python;
+		using namespace knowrob;
+
+		/////////////////////////////////////////////////////
+		// mappings for static functions
+		def("InitKnowledgeBaseWithArgs", &InitKnowledgeBaseWrapper, "Initialize the Knowledge Base with arguments.");
+		def("InitKnowRob", &InitKnowledgeBaseFromSysArgv, "Initialize the Knowledge Base using sys.argv.");
```
daniel86: here is a minor name missmatch

daniel86 (approved):

PR #406, opened 2024-07-31
Propertytree handling in Python

sasjonge:
With this PR i improved the Propertytree handling in python, and also allowed the loading from settings from a python dict.


include/knowrob/PropertyTree.h
```diff
@@ -28,19 +28,37 @@ namespace knowrob {
 		 */
 		explicit PropertyTree(const boost::property_tree::ptree *ptree);
 
+		/**
+		 * Load a reasoner configuration from a JSON string.
+		 * @param json_str a JSON string.
+		 */
+		explicit PropertyTree(const std::string &json_str);
```
daniel86: minor comment: usually it is better to use string_view. Reason: you can call the function then either with string or string_view (string can implicitly be casted to string_view but not the other way around!)

src/PropertyTree.cpp
```diff
@@ -18,15 +22,24 @@ PropertyTree::PropertyTree()
...
-	for (const auto &key_val: *ptree) {
-		loadProperty(key_val.first, key_val.second);
-	}
+PropertyTree::PropertyTree(const std::string &json_str)
+		: PropertyTree() {
+	std::istringstream ss(json_str);
+	boost::property_tree::ptree tree;
+	boost::property_tree::read_json(ss, tree);
+	// assign the variable with a new memory allocation
+	ptree_ = new boost::property_tree::ptree(tree);
```
daniel86: there is a memory leak here because no one calls **delete ptree_**.
You can also not just call delete in the destructor because in the case above PropertyTree *is not the owner*. So you could add another flag indicating whether delete must be called in the destructor.

src/ontologies/DataSource.cpp
```diff
@@ -159,5 +159,7 @@ namespace knowrob::py {
 				.staticmethod("getVersionFromURI")
 				.def("create", &DataSource::create)
 				.staticmethod("create");
+		// register shared pointer
+		register_ptr_to_python< std::shared_ptr< DataSource > >();
```
daniel86: so this is really needed?

daniel86 (changes requested):
Looks good! but there is a memory leak that should be fied

src/PropertyTree.cpp
sasjonge: i changed ptree_ now to be a smart pointer

include/knowrob/PropertyTree.h
```diff
@@ -26,7 +26,18 @@ namespace knowrob {
...
 		 * @param ptree a property tree.
 		 */
-		explicit PropertyTree(const boost::property_tree::ptree *ptree);
+		explicit PropertyTree(std::shared_ptr<const boost::property_tree::ptree> ptree);
+
+		/**
+		 * Load a reasoner configuration from a JSON string.
+		 * @param json_str a JSON string.
+		 */
+		explicit PropertyTree(const std::string_view &json_str);
```
daniel86: minor: You should not use by reference for string_view, always pass string_view by value

@see https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

daniel86 (approved):

PR #407, opened 2024-08-01
Work on GoalDrivenReasoner interface

daniel86:
The interface is simplified and renamed from `submitQuery` to `evaluateQuery`.
Main points are:
- reasoner do not need to deal with `TokenBuffer` etc, and they do not need to send a STOP token to indicate they are done
- the threading is now done centrally such that every reasoner runs in a worker thread


sasjonge (approved):
Looks good to me.

PR #408, opened 2024-08-05
Revised interface for GoalDrivenReasoner

daniel86:
- A new type `SimpleConjunction` (a conjunction of literals) is introduced
- The `ReasonerQuery` type accepted by `GoalDrivenReasoner` was changed to use `SimpleConjunction` instead of a single literal
- A testcase is added for PrologReasoner that tests if `SimpleConjunction` can be handled through the `evaluateQuery` interface
- I encountered a bug with unsigned type which is also fixed by this PR

Note that the `SupportsSimpleConjunctions` feature is currently ignored by the knowledge base. A consecutive PR could address this and make use of the feature flag in query pipeline construction.


sasjonge (approved):
Looks good.

PR #409, opened 2024-08-05
Fix python kernel crash in Jupyter Notebooks

sasjonge:
In Jupyter Notebooks we regularly get this error:

Fatal Python error: PyThreadState_Get: the function must be called with the GIL held, but the GIL is released (the current Python thread state is NULL)

This seems a threading issue, in the init function of Knowrob we call Py_Initialize, which could be a problem because a python interpreter is created inside of a running python interpreter. To fix this I sourround the call with Py_Initialized. 

This seems to fix the issue.


daniel86 (approved):
seems unlikely, see here https://docs.python.org/2/c-api/init.html#c.Py_Initialize

"This is a no-op when called for a second time".

But does definetly not hurt to merge

PR #433, opened 2024-09-03
SOMA_DFL GoalReasoner

mpomarlan:
WiP, to do:

1) config file 
2) test cases
3) (upstream changes) SOMA_DFL refactoring: folder structure to make all OWLs and resource files attached to the python package, and a class object for reasoning config and cache.

Note, this will require a pip install from a https://github.com/ease-crc/ease_lexical_resources clone to work, because the reasoner depends on the dfl package.

### DIFF
--- a/src/reasoner/soma_dfl/soma_dfl.py
+++ b/src/reasoner/soma_dfl/soma_dfl.py
@@ -1,0 +1,144 @@
+from knowrob import *
+
+import dfl.semrep as sr
+
+def _iriOrVariable(x):
+    if x.isVariable():
+        return x
+    return IRIAtom(str(x))
+
+class SOMADFLReasoner(GoalReasoner):
+    def __init__(self):
+        super().__init__()
+        self.hasDisposition = IRIAtom("http://ease-crc.org/ont/SOMA_DFL.owl#hasDisposition")
+        self.isDispositionOf = IRIAtom("http://ease-crc.org/ont/SOMA_DFL.owl#isDispositionOf")
+        self.hasPart = IRIAtom("http://ease-crc.org/ont/SOMA_DFL.owl#hasPart")
+        self.isPartOf = IRIAtom("http://ease-crc.org/ont/SOMA_DFL.owl#isPartOf")
+        self.hasConstituent = IRIAtom("http://ease-crc.org/ont/SOMA_DFL.owl#hasConstituent")
+        self.isConstituentOf = IRIAtom("http://ease-crc.org/ont/SOMA_DFL.owl#isConstituentOf")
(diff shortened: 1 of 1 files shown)


src/reasoner/soma_dfl/soma_dfl.py
```diff
@@ -0,0 +1,144 @@
...
+                query_term = GraphSequence([GraphPattern(Variable("z"), self.rdfType, IRIAtom(sc))])
+                self.storage().query(GraphQuery(query_term), lambda bindings : retq.add(str(bindings.get(Variable("z")))))
+            retq = list(retq)
+        return retq
+                
+    def initializeReasoner(self, config: PropertyTree) -> bool:
+        # TODO: MAJOR: separate cache, namespace etc. for each instance of this class, right now it is only allowed to be a singleton
+        # TODO: parse and use config.
+        # Flags that would make sense:
+        #     ontology and usematch files to use
```
daniel86, line 76: there is a mechanism dedicated for loading input files into a reasoner (and not into the storage backend)

In settings files, you can have an import section like here:
https://github.com/knowrob/knowrob/blob/dev/tests/settings/lpn.json#L28

**"format": "prolog"** is important. Now in the reasoner, you can tell KnowRob about that your reasoner can handle specific formats like the Prolog reasoner does:
```
	addDataHandler("prolog", [this]
			(const DataSourcePtr &dataFile) { return consult(dataFile->uri()); });
```

https://github.com/knowrob/knowrob/blob/dev/src/reasoner/prolog/PrologReasoner.cpp#L89

daniel86: Might not be covered in unit tests in the pre-release, so could be that the GIL crashes when you try from Python. **edit** should work in dev branch

daniel86:
looks good to me! A few things I spotted where the API could be better on the KnowRob side, but well it is not overly verbose I would say so seems good.

mpomarlan:
Item 3 on the todos is done, ease_lexical_resources is refactored.

Will also push this towards Sascha's notebook/Giang's cloud setup.

sasjonge:
@mpomarlan Hey, can this PR be closed? I think we used this specialized code in another repo. If I understand it correctly, in the future a more general (and then potentially Data Driven Reasoner) is planned?

mpomarlan:
We could also try to merge this into knowrob, depending on what Daniel says, but if we do we need to match the code here with that in your notebook, for bugfixes.

Eventually a data driven reasoner is planned for a class of languages, as opposed to specific queries on a specific ontology, so that of course will be of more general use.

PR #446, opened 2024-09-18
Added unit testing for notebooks and documentation for tutorials

sasjonge:


tests/py/test-jupyter.cpp
```diff
@@ -21,4 +21,4 @@ class JupyterTests : public PythonTests {
 #define TEST_JUPYTER(notebook) EXPECT_NO_THROW(\
         PYTHON_TEST_CALL0("test_notebook", python::object(URI::resolve(notebook))));
 
-TEST_F(JupyterTests, python_kb) { TEST_JUPYTER("jupyter/intro.ipynb"); }
+TEST_F(JupyterTests, python_kb) { TEST_JUPYTER("tutorials/intro.ipynb"); }
```
daniel86, line 24: do we need this test at all still? if it is redundant with the python testing, I think we could drop the JupyterTests class

tutorials/notebook_test_util.py
```diff
@@ -0,0 +1,119 @@
+import nbformat
+import os
+import shutil
+import sys
+import argparse 
+
```
daniel86, line 6: so to make it easier to maintain this file, I think it should receive a description of what it is used for.

daniel86 (approved):
Looks good!

But can we somehow teach gtest to run this? Or at least some target for make. Because we need to have an easy way for devs to run the tests locally in my opinion. It would be best if we could still run "all tests" in one command. Maybe there is a way to teach test to run the Jupyter tests too?

Else maybe we should keep the redundancy for now that we also run them through gtest such that devs can quickly see if no exception occurs

daniel86:
an advantage of gtest is that we get annotations in github CI! (but not with the current implementation though, as we parse the json file in the gtest at the moment)

but maybe there is also something available for the testing framework you make use of for the notebooks?

PR #454, opened 2025-03-19
Bugfix: handling of duplicate variables in MongoDB triple patterns

### LINKED ISSUE #448: Using variables multiple times with the MongoDB backend is broken
Queries where one Variable is used multiple times (e.g. `rdf:type(A,A) `) result in an error message when run via the mongolog backend.

Example in the knowrob terminal on commit 787875e44640d64b9fd11d1bebd9a10c0f1627ce:
```
$ ./knowrob-terminal --config-file ../tests/settings/mongolog-lpn.json
[14:25:39.743] [info] [KnowRob] static initialization done.
[14:25:39.743] [info] [KnowRob] Writing logs to /home/teddy/.knowrob/logs
Welcome to KnowRob.
For online help and background, visit http://knowrob.org/

?- rdf:type(A,A) 
no.
[14:25:45.945] [warning] an exception occurred for graph query ((rdf:type(A,A))): BackendError: [mongo] cursor_error: Invalid $set :: caused by :: specification contains two conflicting paths. Cannot specify both 'v_VARS.A.val' and 'v_VARS.A.val': { v_VARS.A.val: { $cond: [ { $not: [ "$v_VARS.A.val" ] }, "$next.s", "$v_VARS.A.val" ] }, v_VARS.A.val: { $cond: [ { $not: [ "$v_VARS.A.val" ] }, "$next.o", "$v_VARS.A.val" ] } }..
```

PS: The named commit is on my fork and introduces a test case for this. If you want, i can make a PR for that.

daniel86:
This PR fixes a bug with the MongoDB storage backend that occurs when a variable appears at multiple positions in a triple pattern as in `triple(?x, test:b, ?x)`. Solution is to only $set the first variable value, then to use $match to ensure the value is the same as for the duplicate variable.

Closes #448


sasjonge (approved):
Looks good!

PR #461, opened 2025-05-03
Build static

HoangGiang93:
**Hola 🙂**

I'm glad to share that I've finally built a Python version of KnowRob with *minimal dependencies*—in fact, it only requires a single shared library: `libboost_python<version>.so`.

Since each Python version needs its own build, I’ve prepared [precompiled packages](https://nc.uni-bremen.de/index.php/s/ynPX3skq4iYBCZt) for three versions:

* **Python 3.8** (for Ubuntu 20.04)
* **Python 3.10** (for Ubuntu 22.04)
* **Python 3.12** (for Ubuntu 24.04)

Each package includes the corresponding `libboost_python.so.1.88.0`, and all builds were done on Ubuntu 20.04 to ensure compatibility with Ubuntu 20.04 and newer.

In the attached video, you'll see just how easy it is to use the KnowRob Python library on Ubuntu ≥ 20.04. Here’s a quick guide:

[Thumbnail](https://nc.uni-bremen.de/index.php/s/wwbnb6KyyRa5bwL)

---

### 🐳 How to run it in Docker:

1. **Start a Docker container**

   ```bash
   docker run -it ubuntu:2x.04 bash  # Replace x with 0, 2, or 4
   apt update && apt install -y software-properties-common
   add-apt-repository ppa:ubuntu-toolchain-r/test
   apt install -y g++-11 gcc-11 libpython3-dev binutils
   ```

   > This ensures `GLIBCXX ≥ 3.4.30` and Python 3 is available.

2. **Copy the `dist-packages` folder into the container**

   ```bash
   docker cp <path/to/dist-packages> <container_id>:/
   ```

3. **Use the KnowRob Python library**

   ```bash
   cd dist-packages
   export LD_LIBRARY_PATH=.  # Ensure libboost is found
   python3  # Not just `python`
   >>> from knowrob import *
   ```

---

**TL;DR**:
KnowRob now works out-of-the-box as [static libraries](https://github.com/Multiverse-Framework/Multiverse-Knowledge/tree/KnowRobIntegration/knowrob_connector/lib) without complex installation steps or dependency issues. Just download, extract, and import. All libraries are the latest stable versions (see [setup.sh](https://github.com/Multiverse-Framework/Multiverse-Knowledge/blob/KnowRobIntegration/setup.sh)).

Python: https://nc.uni-bremen.de/index.php/s/ynPX3skq4iYBCZt
C++: https://nc.uni-bremen.de/index.php/s/xazXWE8E5i2rjec (only static libs)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,11 +324,17 @@ set(KNOWROB_LIBRARIES
         ${GTEST_LIBRARIES})
 target_link_libraries(knowrob ${KNOWROB_LIBRARIES})
 # install shared library and headers
-install(TARGETS knowrob
-        PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_PREFIX}/knowrob)
-install(FILES $<TARGET_FILE:knowrob>
-        DESTINATION ${CMAKE_INSTALL_PREFIX}/lib
-        RENAME libknowrob${PYTHON3_SUFFIX})
+
+option(KNOWROB_USE_SUFFIX "Append fixed suffix to KnowRob module name" OFF)
+if(KNOWROB_USE_SUFFIX)
+	install(FILES $<TARGET_FILE:knowrob>
+			DESTINATION ${CMAKE_INSTALL_PREFIX}/lib
+			RENAME libknowrob${PYTHON3_SUFFIX})
+else()
+	install(TARGETS knowrob
+			PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_PREFIX}/knowrob)
+endif()
+
 install(DIRECTORY include/knowrob
         DESTINATION include)
 # Install Python and Prolog files into the share directory such that they can be loaded at runtime.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,7 +328,8 @@ target_link_libraries(knowrob ${KNOWROB_LIBRARIES})
 
 if(KNOWROB_USE_SUFFIX)
 	message(STATUS "Installing knowrob with suffix ${PYTHON3_SUFFIX}")
-	set_target_properties(knowrob PROPERTIES OUTPUT_NAME "knowrob${PYTHON3_SUFFIX}")
+	string(REPLACE ".so" "" PYTHON3_SUFFIX_BASE "${PYTHON3_SUFFIX}")
+	set_target_properties(knowrob PROPERTIES OUTPUT_NAME "knowrob${PYTHON3_SUFFIX_BASE}")
 endif()
 install(TARGETS knowrob
 		PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_PREFIX}/knowrob)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 message(STATUS "Found Python3 include: ${Python3_INCLUDE_DIRS}")
 message(STATUS "Found Python3 libraries: ${Python3_LIBRARIES}")
 execute_process(
-    COMMAND python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}-config --extension-suffix
+    COMMAND ${Python3_EXECUTABLE}-config --extension-suffix
     OUTPUT_VARIABLE PYTHON3_SUFFIX
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
@@ -134,7 +134,7 @@ if (CATKIN)
         catkin_package(
                 INCLUDE_DIRS include
                 LIBRARIES knowrob
-                DEPENDS Boost -lswipl MONGOC RAPTOR REDLAND spdlog Python3 GTest EIGEN3
+                DEPENDS Boost -lswipl MONGOC RAPTOR2 REDLAND spdlog Python3 GTest EIGEN3
         )
     else ()
         message(STATUS "No catkin workspace found.")
@@ -152,12 +152,13 @@ set(knowrob_INCLUDE_DIRS
         ${Boost_INCLUDE_DIRS}
         ${Python3_INCLUDE_DIRS}
         ${MONGOC_INCLUDE_DIRS}
-        ${RAPTOR_INCLUDE_DIRS}
+        ${RAPTOR2_INCLUDE_DIRS}
         ${REDLAND_INCLUDE_DIRS})
 include_directories(${knowrob_INCLUDE_DIRS})
 
 option(KNOWROB_USE_SUFFIX "Append fixed suffix to KnowRob module name" OFF)
-add_library(knowrob SHARED
+set(KNOWROB_LIBRARY_TYPE STATIC CACHE STRING "Library type")
+add_library(knowrob ${KNOWROB_LIBRARY_TYPE}
         src/knowrob.cpp
         src/ThreadPool.cpp
         src/Logger.cpp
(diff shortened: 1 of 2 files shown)


daniel86:
@HoangGiang93 Hola, it seems merging your other request caused a merge conflict with this one...

HoangGiang93:
> @HoangGiang93 Hola, it seems merging your other request caused a merge conflict with this one...

That's weird :/ This branch is actually built on top of the other one :/
It starts from the commit https://github.com/knowrob/knowrob/pull/461/commits/7efb59a8c9121ee7cbe698a651dcafb08a42a5da

include/knowrob/integration/python/converter/dict.h
```diff
@@ -10,7 +10,7 @@
 namespace knowrob::py {
 
 	// Function to convert Python object to boost::any
-	boost::any python_to_boost_any(const boost::python::object &obj) {
+	inline boost::any python_to_boost_any(const boost::python::object &obj) {
```
daniel86, line 13: Is there a strong reason for inlining this functions? I mean, shouldn't hurt, just wondering why you did change that

CMakeLists.txt
```diff
@@ -1,39 +1,46 @@
...
+
 # Make sure some specific warnings are enabled
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wswitch -Wno-undef -Wnon-virtual-dtor")
 set(CMAKE_CXX_STANDARD 17)
-if (POLICY CMP0148)
-    cmake_policy(SET CMP0148 OLD)
-endif ()
+
+if(POLICY CMP0148)
+	cmake_policy(SET CMP0148 OLD)
```
daniel86, line 10: It would be good if you could avoid these auto-formatting changes in pull requests as they make it much harder for me to review!

CMakeLists.txt
```diff
@@ -43,418 +50,536 @@ include("cmake/ontologies.cmake")
...
+	pkg_check_modules(CURL REQUIRED libcurl)
+	pkg_check_modules(SSL REQUIRED libssl)
+	pkg_check_modules(CRYPTO REQUIRED libcrypto)
+	pkg_check_modules(GMP REQUIRED gmp)
+	pkg_check_modules(YAJL REQUIRED yajl)
+	pkg_check_modules(UUID REQUIRED uuid)
+	pkg_check_modules(LDAP REQUIRED ldap)
+	pkg_check_modules(LBER REQUIRED lber)
+	pkg_check_modules(PCRE REQUIRED libpcre)
+	link_directories(${MONGOC_DIR}/../../lib/
```
daniel86, line 368: why the relative path "../../lib" ? seems like a hack to me

CMakeLists.txt
```diff
@@ -43,418 +50,536 @@ include("cmake/ontologies.cmake")
...
+		ltdl
+		tinfo
+		fmt
+		dl
+		z
+		rt)
+	message(STATUS "KnowRob uses libraries: ${KNOWROB_LIBRARIES}")
+endif()
+
+set(OpenSSL_LIBRARIES "") # Unset to avoid leakage
```
daniel86, line 416: what leakage and where?

CMakeLists.txt
```diff
@@ -43,418 +50,536 @@ include("cmake/ontologies.cmake")
...
 target_link_libraries(all_gtests
-        -Wl,--whole-archive,--no-as-needed
-        knowrob
-        -Wl,--no-whole-archive
-        ${Boost_Python_COMPONENT}
-        ${GTEST_MAIN_LIBRARIES})
-
+	-Wl,--whole-archive,--no-as-needed
+	knowrob
+	libswipl.so # Gtest needs this to find the Prolog engine
```
daniel86, line 582: this is not true for me locally, nor in CI! gtest runs just fine without

daniel86 (changes requested):
There are a few points that should be clarified before merging (see my comments).

In addition there is one major concern regarding violation of licenses. Afaik there are more constraints on distributing statically linked libraries. E.g. knowrob is BSD, so we are not allowed to distribute any statically linked GPL library, right? In any case we can allow people to build their own statically linked library, but as far as I can see it makes only sense in case that we distribute it (i.e. via github releases, and maybe later on via pip). Did you consider licensing issues? I think we would need to check license for each statically linked library....

daniel86:
> > @HoangGiang93 Hola, it seems merging your other request caused a merge conflict with this one...
> 
> That's weird :/ This branch is actually built on top of the other one :/ It starts from the commit [7efb59a](https://github.com/knowrob/knowrob/commit/7efb59a8c9121ee7cbe698a651dcafb08a42a5da)

maybe because I did merge via squash -- I kind of prefer squash and having a clean change log, also good for generating release notes

include/knowrob/integration/python/converter/dict.h
HoangGiang93: Yes it hurts, when I build static libraries, it's expected to have a library as the implementation of a header file. In this case they are the same so it wouldn't work.

CMakeLists.txt
HoangGiang93: This was from your original CMakeLists.txt

CMakeLists.txt
HoangGiang93: without it I couldn't run the test for static libraries. In CI you have libswipl.so already in knowrob, that's why it works

CMakeLists.txt
HoangGiang93: yes, there is a bug that always makes MONGO_LIBRARY_DIRS not the way it should be :/ I don't know why :/ it only happens with mongoc

CMakeLists.txt
daniel86: no, it is not. there is no mention of "OpenSSL_LIBRARIES" in the current cmake

include/knowrob/integration/python/converter/dict.h
daniel86: ok fine

CMakeLists.txt
HoangGiang93: Then I'm confused too :confused: I can remove it

CMakeLists.txt
HoangGiang93: sure :) but it only happens once :))

CMakeLists.txt
daniel86: I thought it maybe was related to the problem we have with SSL plus raptor? i.e. the issue with gazebo. But at least in knowrob we do not use the library

CMakeLists.txt
daniel86: fair enough :) maybe we should add an entry to https://github.com/knowrob/knowrob/blob/dev/.editorconfig about CMake files? could be good to avoid ping-ponging between different conventions.

CMakeLists.txt
daniel86: but you did include swipl in the list of statically linked libraries, shouldn't that be enough? to me it seems we link swipl two times into `all_gtests` one time statically through `knowrob` target and one time dynamically

HoangGiang93: yes that should work normally. But because for some reason it needs some more functions in runtime :( for that the static libraries aren't enough :/

HoangGiang93:
> There are a few points that should be clarified before merging (see my comments).
> 
> In addition there is one major concern regarding violation of licenses. Afaik there are more constraints on distributing statically linked libraries. E.g. knowrob is BSD, so we are not allowed to distribute any statically linked GPL library, right? In any case we can allow people to build their own statically linked library, but as far as I can see it makes only sense in case that we distribute it (i.e. via github releases, and maybe later on via pip). Did you consider licensing issues? I think we would need to check license for each statically linked library....

About licensing I have no idea, in my [setup.sh](https://github.com/Multiverse-Framework/Multiverse-Knowledge/blob/KnowRobIntegration/setup.sh) I put instructions of how to build it so one could see how we build it.

HoangGiang93:
Let me know if you need any support from my side, I will be on vacation the whole next two weeks.

daniel86:
@HoangGiang93 hey, sorry was busy with demo preparation. Also I thought you will update the pull request next? i.e. merging dev such that we do not have a merge conflict anymore?
Also not all of the discussion points are resolved, e.g. the meaningless line about SSL should be removed.

I'll have a look at the strange part about mongo then...

PR #462, opened 2025-05-16
Exporting triples to file

daniel86:
This PR adds an interface to `KnowledgeBase` and `QueryableStorage` to export all triples in the model to a file. `RDF/XML` and `Turtle` format are supported for export.


sasjonge:
Tested it and result looks good!

